PR #353, opened 2019-02-21
Fix cmake bug and remove unused variable

david-koenig:
It is possible for git describe command in GitRevision.cmake to
return a success error code but also an empty string for its output
when it cannot find a git commit. In this case code was hitting
a syntax error in the regex on line 35, because cmake thought the
string command had five arguments instead of six.

This syntax error could be fixed by placing quotation marks around the
last argument, but this would be a logical error. An empty string here
should be treated the same way as an error exit code, meaning do not
append a git version hash string to the "vX.Y.Z" version string in the
user agent.

This code change also unrelatedly removes one unused variable from a test
that was causing a warning message.


david-koenig:
CBMC failure is a known issue that is being fixed in another PR

PR #359, opened 2019-04-29
Session from keyring

david-koenig:
This adds a new API function to create sessions directly from keyrings, using a default CMM under the hood. It also modifies all of the existing examples, other than the caching CMM example, to use the new function and not to talk about CMMs.


david-koenig:
I've just done a further update to add a ``aws_cryptosdk_caching_cmm_new_from_keyring`` function which similarly shortcuts the creation of the default CMM for the user.

I also renamed ``aws_cryptosdk_caching_cmm_new`` to ``aws_cryptosdk_caching_cmm_new_from_cmm`` to better match the naming conventions of the other functions. Note that this is a breaking change, but this seems to be the best time to do it, while we are still in preview.

examples/kms_discovery.cpp
```diff
@@ -14,11 +14,15 @@
  */
 
 #include <aws/cryptosdk/cpp/kms_keyring.h>
-#include <aws/cryptosdk/default_cmm.h>
 #include <aws/cryptosdk/session.h>
 
-const char *KEY_ARN_US_WEST_2    = "arn:aws:kms:us-west-2:658956600833:key/b3537ef1-d8dc-4780-9f5a-55776cbb2f7f";
-const char *KEY_ARN_EU_CENTRAL_1 = "arn:aws:kms:eu-central-1:658956600833:key/75414c93-5285-4b57-99c9-30c1cf0a22c2";
+/* This example shows how various KMS keyrings operate slightly differently on
+ * decryption. The code is completely functional except for the two fake KMS
```
ttjsu-aws: Do we want to document the difference in operation that is being referred to here? It is not clear to me what the comment is trying to convey.

examples/multi_keyring.cpp
```diff
@@ -48,9 +43,7 @@ void encrypt_or_decrypt(
     if (!aws_cryptosdk_session_is_done(session)) abort();
     if (input_consumed != input_len) abort();
 
-    /* This destroys both the session and the CMM, but not the keyring, since the
-     * keyring pointer was not released.
-     */
+    /* This destroys both the session but not the keyring, since the keyring pointer was not released. */
```
ttjsu-aws: "This destroys the session and not the keyring.."

include/aws/cryptosdk/cache.h
```diff
@@ -531,14 +531,42 @@ AWS_CRYPTOSDK_STATIC_INLINE void aws_cryptosdk_materials_cache_release(
  * @param cache_limit_ttl_units The units of cache_limit_ttl. Allowable values are defined in aws/common/clock.h
  */
 AWS_CRYPTOSDK_API
-struct aws_cryptosdk_cmm *aws_cryptosdk_caching_cmm_new(
+struct aws_cryptosdk_cmm *aws_cryptosdk_caching_cmm_new_from_cmm(
```
ttjsu-aws: I find the name `caching_cmm_new_from_cmm` to be slightly confusing, can we be more specific?

include/aws/cryptosdk/cache.h
```diff
@@ -531,14 +531,42 @@ AWS_CRYPTOSDK_STATIC_INLINE void aws_cryptosdk_materials_cache_release(
...
     const struct aws_byte_buf *partition_id,
     uint64_t cache_limit_ttl,
     enum aws_timestamp_unit cache_limit_ttl_units);
 
+/**
+ * Creates a new instance of the caching crypto materials manager. This CMM will intercept requests for encrypt
+ * and decrypt materials, and forward them to the provided materials cache.
+ *
+ * If multiple caching CMMs are attached to the same materials_cache, they will share entries if and only if the
+ * partition_id parameter is set to the same string. Unless you need to use this feature, we recommend passing NULL, in
```
ttjsu-aws: minor: This line seem to be overflowing github viewing limit, could you run the style-formatting script to limit the number of characters to 120?

include/aws/cryptosdk/session.h
```diff
@@ -56,6 +56,22 @@ struct aws_cryptosdk_session;
 
 enum aws_cryptosdk_mode { AWS_CRYPTOSDK_ENCRYPT = 0x9000, AWS_CRYPTOSDK_DECRYPT = 0x9001 };
 
+/**
+ * Creates a new encryption or decryption session.
+ *
+ * @return The new session, or NULL on failure (in which case, an AWS error code is set)
```
ttjsu-aws: "The new session object.."

include/aws/cryptosdk/session.h
```diff
@@ -56,6 +56,22 @@ struct aws_cryptosdk_session;
...
+ *
+ * @param allocator The allocator to use for the session object and any temporary
+ *                  data allocated for the session
+ * @param mode The mode (AWS_CRYPTOSDK_ENCRYPT or AWS_CRYPTOSDK_DECRYPT) to start
+ *             in. This can be changed later with @ref aws_cryptosdk_session_reset
+ * @param keyring The keyring which will encrypt or decrypt data keys for this session.
+ *                This function uses a default CMM to link the session and keyring.
```
ttjsu-aws: If we are making the `default_cmm` to be implemented under the hood/ a private implementation detail, I dont think we should mention that in the public APIs documentation here.  This goes back to my lack of complete understanding on what is the motivation for this change.

examples/string.cpp
```diff
@@ -14,7 +14,6 @@
  */
 
 #include <aws/cryptosdk/cpp/kms_keyring.h>
-#include <aws/cryptosdk/default_cmm.h>
```
ttjsu-aws: Since the contents of `default_cmm.h` is now a private implementation detail we should consider moving this header file from `include/default_cmm.h` to `include/private/default_cmm.h`.

examples/kms_discovery.cpp
david-koenig: The comment was trying to give an overall context of what this example is about. It's not a realistic one because no one needs to make six different keyrings to do the exact same decryption. It's meant to be illustrative of what the differences in behavior of those different keyrings are through the comments lower down in the file. I'll reword this to make it a little clearer.

examples/multi_keyring.cpp
david-koenig: Good catch. I missed that.

include/aws/cryptosdk/cache.h
david-koenig: So C doesn't have classes and objects, but we are essentially simulating a C++ or Java-like behavior here by making functions that are effectively constructors. In C++ or Java, the functions ``aws_cryptosdk_caching_cmm_new_from_cmm`` and ``aws_cryptosdk_caching_cmm_new_from_keyring`` would both be constructors named simply ``caching_cmm``. Then you would allocate a new instance on the heap using the ``new`` keyword with ``new caching_cmm(cmm)`` or ``new caching_cmm(keyring)``, i.e., You would define two different constructors to take either a CMM or keyring as the argument.

However, there are some significant differences in the C language which force us to do things differently.

First, C++ has namespaces and Java has packages, both of which shield the function names from the general namespace and stop our function names from colliding with userspace function names. C has no such thing, so we are making sure that all public names in this SDK start with ``aws_cryptosdk_`` to avoid those collisions.

Second, C has no ``new`` keyword, and what is doing the construction is just a regular function call, so we include the word ``new`` in the function name.

Third, C does not allow you to override the same function name to take arguments of different types the way that C++ and Java do, so you must use different function names in order to take different argument types. For both the session constructors and the caching CMM constructors, the way we are doing this is by using the suffix ``from_cmm`` or ``from_keyring`` to indicate which argument the function takes.

Fourth, when we build up a function name from many constituent parts in this way, it is good practice to put the most general things first and gradually move to the more specific later in the function name. This assures that different functions that are related have the same prefix and makes their grouping more obvious when they are used near each other in the code.

Fifth, both of these functions would only need to take the CMM or the keyring as the argument in C++ or Java, but since C does explicit memory management and aws-c-common has abstracted this into the allocator object, which we need to pass around.

So all of these reasons together are why these construction calls become ``aws_cryptosdk_caching_cmm_new_from_cmm(allocator, cmm)`` and ``aws_cryptosdk_caching_cmm_new_from_keyring(allocator, keyring)``. My opinion is that these names are exactly as specific as they should be, and I don't want to make them any longer than they have already gotten. I hope this explains clearly how we got to these names.

include/aws/cryptosdk/cache.h
david-koenig: I agree that these particular lines are a bit long, but I literally copied the exact same lines from the other caching CMM constructor documentation and just modified the description of the one parameter that is different.

I'm fine with modifying the clang-format settings to change the allowed line lengths, but that will potentially affect a lot more files throughout the code base, so it should be done in a separate code review.

Note that all of these document comments will be rendered in the doxygen web pages. I need to put a PR up for that separately because it goes into another branch of the repo. But when I put that PR up after this one, you will be able to read the new wording added here more easily on the webpages, so you'll have a second chance to review the wording here if you want to change something.

I'll manually change the lengths of these lines this one time to make them easier to read, but in the future you can also just pull my branch into your local git repo to read it there rather than on github's browser interface.

include/aws/cryptosdk/session.h
david-koenig: Calling it a session object is not consistent with the terminology we have been using throughout the API documentation. Note that this is not new wording in this PR. I copied the exact wording already used in the other session constructor and just modified the description of the single changed parameter.

include/aws/cryptosdk/session.h
david-koenig: We are not making the ``default_cmm`` a private implementation detail. We're taking nothing out of the public API. We are providing convenience functions so that users don't have to worry about the details of the default CMM if they don't want to.

We need to keep the CMM interface public, because we still want users to have the ability to create their own CMMs and plug whichever ones they want into sessions.

Furthermore, if users want to change the algorithm suite from the default one, they currently need to manually construct a default CMM to do that. Most users won't need this functionality, but we're not going to take it away.

(Note that we could possibly add functionality to set the algorithm ID in the session, but that would only be for a single encryption or decryption. Some users might still want the functionality to have a CMM persistent across many encryptions and decryptions to always use one of the alternative algorithm suites.)

The tl;dr here is that many users don't need to worry about default CMMs and we're giving them a simpler interface where they can just create a keyring and session and never think about CMMs. But power users still need access to the CMMs.

examples/string.cpp
david-koenig: See above comment. This is not a private implementation detail. Default CMMs are still part of the public API, but fewer people need to use them.

CHANGELOG.rst
```diff
@@ -16,3 +16,9 @@ Changelog
 * Fixed empty string bug on git version of KMS user agent
 * Local tests only by default
 * Fix of MAP_ANONYMOUS issue for older Linuxes
+
+0.1.3 -- 2019-04-29
+===================
+* Added API function to make session from keyring
+* Added API function to make caching CMM from keyring
+* Added "_from_cmm" to end of "aws_cryptosdk_caching_cmm_new" function name
```
mattsb42-aws: We are removing an API surface. This is a breaking change, and thus per our verisoning policy this must be a minor version bump (ie: `0.2.0`).

include/aws/cryptosdk/cache.h
ttjsu-aws: Gotcha, thank-you for the explanation. I think I was thinking more on the lines of do we want to explicitly state `upstream cmm`. But on second thought, this seems fine to me.

include/aws/cryptosdk/cache.h
ttjsu-aws: I didnot notice that the prior lines also had this issue, we should probably make the change to those lines as well to be consistent.

include/aws/cryptosdk/session.h
ttjsu-aws: The reason I suggested `session object` is that the in trailing lines we use that terminology.

include/aws/cryptosdk/session.h
ttjsu-aws: Alright, thank-you for the explanation.

examples/string.cpp
ttjsu-aws: Gotcha!

CHANGELOG.rst
david-koenig: Okay, I didn't think these issues necessarily applied to the pre 1.0.0 versions, but it's no big deal to switch this to 0.2.0. I will make that change and then merge the PR.

david-koenig: Ah, @mattsb42-aws I noticed that you clicked the button to block the merge on your requested change, so I will let you review the version bump. Please feel free to merge this code yourself if everything is cool.

I decided that it was a good time to get one more name change in, since we're making a minor version bump anyway and it's better to get things stable before we put out the 1.0.0 release.

So I'm adding another commit to change the function ``aws_cryptosdk_session_get_algorithm`` to ``aws_cryptosdk_session_get_alg_id``. This is more shorter, more accurate, and more consistent with the other public API function we have that refers to algorithm IDs, i.e., ``aws_cryptosdk_default_cmm_set_alg_id``

Everything else is the same as the code that @ttjsu-aws already approved.

PR #360, opened 2019-04-30
Doc update for v0.2

david-koenig:
two new constructors for session and caching CMM from keyrings

This is the companion doc code change for PR #359 . Pull the branch to your local box and do Open File on html/index.html on your web browser to see the rendered pages.


david-koenig:
This has been updated to include the change of ``aws_cryptosdk_session_get_algorithm`` to ``aws_cryptosdk_session_get_alg_id`` and the version number at the top of the API docs is now listed as "0.2.x".

It seems to be a good idea not to specify the patch number on the public API documents so that in the future when we make patch changes that don't alter the public API, the version on the public docs won't look out of date.

So for example, let's say we put out v0.2.0 of the code with docs that say "0.2.x" and then we do a few updates to v0.2.1 and v0.2.2 and the API is unchanged, there is no need to republish docs. Then if we added a new function (without breaking any old ones) we could do a doc update that kept the "0.2.x" header at the top, but in the doc comments for the new function we could specify "since 0.2.3". In that way, the documents are still up to date for anyone using the latest major/minor version, regardless of patch number.

ttjsu-aws:
I am not convinced on using the versioning number as 0.2.x on the public API documents, I believe we should be clear to our users on what changes are being made and reflect the version number in the code. From what I understand, making a document update isn't cumbersome and is a quick change. As such, I am not very sure what we are benefiting out of using 0.2.x here.

mattsb42-aws:
I agree with @ttjsu-aws. If a user is looking at the docs, it should not be ambiguous which version(s) they apply to. If we update the version for each patch version, then it is clear that the docs are up-to-date for that patch version and we did not just forget to update them.

This is the same pattern that we follow for all of our other projects. If deploying new builds of the docs is introducing too much friction in the development process, we should invest in automating those builds.

I should also note here that our versioning policy does not preclude modifying the API in patch versions; it simply sets some guidelines for what *size* changes might need a minor version bump and requires that breaking changes *must* happen in a minor version bump.

Side note, once we're past `1.0`, a breaking change like this (namely: removing an API) will require two version changes: one patch version to introduce the new API and deprecate the old API and a subsequent minor version to remove the old API.

david-koenig:
@ttjsu-aws @mattsb42-aws 

Firstly, I agree with Matt's point that post 1.0 we should add a new functionality separately from removing an old one in the case of renaming things.

Secondly, I think the objections to the 0.2.x naming in the API docs was because it seemed imprecise, as if we weren't informing our customers about the exact current version of things. I can see that point of view, because the ".x" looks like a source of uncertainty.

However, I think the most salient point here is that patch version numbers (e.g., the "3" in v1.2.3) most likely should not be printed in the API docs at all. We can do this without using a placeholder by just listing "1.2" or even "v1.2" in the API doc header.

The reason for this is that by our own definition of patch versions from https://github.com/david-koenig/aws-encryption-sdk-c/blob/master/VERSIONING.rst we will not be making API breaking changes when we do patch version changes. So all patch versions v1.2.0 up to v1.2.whatever must have a compatible API.

Furthermore, even if we are making non-breaking changes that are "significant", according to the versioning policy document, those also should get a minor version bump.

The majority of our patch updates are going to have zero impact on the API. They will be under-the-hood bug fixes or improvements in implementation that shouldn't require the publication of new API docs at all. This has already been the case with all of our version bumps prior to #359. The v0.1.0, v0.1.1, and v0.1.2 releases of this library all have identical user APIs, and we never bothered to update the API docs pages, but that means that someone who was using v0.1.1 or v0.1.2 of the code and looked at the API docs would have seen 0.1.0 at the top of those pages and thought they were looking at out of date docs, even though they weren't.

Sure, we could just update the public docs with every non-API changing patch version bump, but why force ourselves to do a bunch of extra busy work, which we will probably get lax about anyway?

Furthermore, this isn't just about trying to save us work, the fact that the document impact of a patch version update will be *at most* the addition of a few "insignificant" functions without any changes to existing functions and documented behavior means that even if we are late getting the doc update out there, the documentation that is in place will still be entirely correct for the latest versions.

In other words, if we have the 3.4.0 docs out there and we do an upgrade to 3.4.1 but get delayed in pushing the 3.4.1 docs out, the documents that are visible to users will still correctly explain things in both the 3.4.0 code and the 3.4.1 code. They will just be missing a few new additions, which we have already deemed are "insignificant."

It seems to me the most logical way to do the API docs in a case like this is as follows: Keep "v3.4" at the top of the doc page, and when there is a function that is only available in a later patch version of that major/minor version combo, we should simply include "(since v3.4.1)" or "(since v3.4.6)" in the comments of that particular function. Because we have deemed that the API changes in these patch versions are going to be insignificant and non-breaking, we shouldn't ever get to the point where we have a very large number of these mid-minor version new API functions, and in this way we will still have accurate documentation for all users of *all* patch versions of our latest major/minor version combo.

This gets at another issue of our Github pages solution for API documents. We don't currently have a way to post documentation for anything older than our latest version of the code. Not everyone is able to upgrade so fast, and some users may want accurate documentation of older versions. The solution I have proposed allows us to at least always have accurate documentation of all patch versions of the latest major-minor combo posted publicly without increasing the complexity of our Github pages document posting mechanism.

Whichever way we go on this, if I'm going to use something other than "0.2.x" at the top of the pages in this PR, I'm going to need to modify the doxygen configuration file in master, which will require another PR in the master branch to do it.

david-koenig:
Regarding @ttjsu-aws's comment "I believe we should be clear to our users on what changes are being made," I agree completely. I am not advocating presenting any ambiguity or imprecision to our users. If you think there is some ambiguity, I believe we're misunderstanding each other.

david-koenig:
So I just pushed another change which just alters the header on the doxygen pages to say "v0.2" so you can see how it looks. If you like this version, then I can make another PR to the master branch to update the doxygen config file so that it actually produces the autogenerated files in this form, but I won't bother with a PR to do that until we reach a consensus on what the document actually should look like.

mattsb42-aws:
> when there is a function that is only available in a later patch version of that major/minor version combo, we should simply include "(since v3.4.1)" or "(since v3.4.6)" in the comments of that particular function

I'll take the blame for not calling this out in the earlier PR, but yes, this is something we need to do for all new additions to the API. This is important regardless of the outcome of this thread, as it serves as an indicator to users as to whether or not they should expect an API component to exist in a version that they are using.

-----

Regarding the `MAJOR.MINOR` vs `MAJOR.MINOR.x` vs `MAJOR.MINOR.PATCH` presentation naming, I know you said that this is not about saving us work, but every reason you give for doing anything other than `MAJOR.MINOR.PATCH` boils down to "it takes extra work to update the docs, so I want to not confuse people if we don't update them".

This does not sound like a reason to make the target of the docs more ambiguous.
This sounds like a reason why we need to prioritize automating the doc updates/builds.

----

Other food for thought: how is a user supposed to know that `0.2` is not actually a specific version that we released at some point without digging though our changelog?

david-koenig:
@mattsb42-aws : This is predicated on the assumption (which I think is reasonable and quite likely) that the vast majority of our patch version bumps are going to involve zero API changes. So the work that you’re saying we need to be disciplined about will be simply making a code update that just modifies a version number in the header. Why put ourselves in the position of having to do that work at all?

I can’t see users going back into our github history of our gh-pages branch to look at old versions of our docs. GitHub doesn’t render those files in any useful way in the history. Anyone who is technically savvy enough to dig through that history and pull up those web pages on their own browser should also be technically savvy enough to do the much better and more direct solution: just build the SDK version they are using with the BUILD_DOCS flag on so they have a local copy of the doxygen pages that they know is for the exact version they are using.

When you see public docs of other open source software projects, do you see those docs listing that they are only for one specific patch version of the code?

david-koenig:
The resolution of our discussion above was to use ``v0.2`` in the header of the doxygen pages and to include a comment on the main page explaining that these docs were generated by ``v0.2.0`` of the code.

I made PR #362 which made those changes and other substantial improvements to version number handling. And I updated this PR with the autogenerated docs which the code in that PR produces.

These should be reviewed together. Pull this branch into your local workspace and open the html/index.html page in your web browser in order to see the results.

PR #377, opened 2019-06-12
Changes in Makefiles and .gitignore to setup infastructure for proofs

angelhof:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #378, opened 2019-06-13
Implement a harness for serialize in framefmt.c

angelhof:
Implement a harness for `aws_cryptosdk_serialize_frame` function in `framefmt.c`. 

Add validity functions for `alg_cryptosdk_alg_properties` and `aws_cryptosdk_frame`, as well as preconditions and postconditions in `aws_cryptosdk_serialize_frame`

Depends on #377 

### DIFF
--- a/.cbmc-batch/include/make_common_data_structures.h
+++ b/.cbmc-batch/include/make_common_data_structures.h
@@ -1,6 +1,3 @@
-#ifndef AWS_CRYPTOSDK_MAKE_COMMON_DATA_STRUCTURES_H
-#define AWS_CRYPTOSDK_MAKE_COMMON_DATA_STRUCTURES_H
-
 /*
  * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
@@ -15,6 +12,8 @@
  * implied. See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+#pragma once
 
 #include <aws/common/common.h>
 #include <aws/cryptosdk/cipher.h>
@@ -39,6 +38,3 @@ void ensure_sig_ctx_has_allocated_members(struct aws_cryptosdk_sig_ctx *ctx);
  * [aws_cryptosdk_alg_props] initializer.
  */
 void ensure_alg_properties_is_allocated(struct aws_cryptosdk_alg_properties **alg_props);
-
-#endif /* AWS_CRYPTOSDK_MAKE_COMMON_DATA_STRUCTURES_H */
-
--- a/.cbmc-batch/include/make_common_data_structures.h
+++ b/.cbmc-batch/include/make_common_data_structures.h
@@ -1,3 +1,6 @@
+#ifndef AWS_CRYPTOSDK_MAKE_COMMON_DATA_STRUCTURES_H
+#define AWS_CRYPTOSDK_MAKE_COMMON_DATA_STRUCTURES_H
+
 /*
  * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
@@ -15,6 +18,7 @@
 
 #include <aws/common/common.h>
 #include <aws/cryptosdk/cipher.h>
+#include <aws/cryptosdk/private/framefmt.h>
 #include <proof_helpers/make_common_data_structures.h>
 #include <proof_helpers/proof_allocators.h>
 #include <stdint.h>
@@ -25,3 +29,16 @@ void ensure_md_context_has_allocated_members(struct aws_cryptosdk_md_context *ct
 
 /* Allocates the members of the context and ensures that internal pointers are pointing to the correct objects. */
 void ensure_sig_ctx_has_allocated_members(struct aws_cryptosdk_sig_ctx *ctx);
+
+/**
+ * This function creates a valid alg properties data structure by
+ * calling the initializer in cipher.h. This is the only valid way to
+ * construct an alg_props structure, and that is why
+ * [aws_cryptosdk_alg_properties_is_valid] checks whether the pointer
+ * of alg_props is the same as the one returned after calling the
+ * [aws_cryptosdk_alg_props] initializer.
+ */
+void ensure_alg_properties_is_allocated(struct aws_cryptosdk_alg_properties **alg_props);
+
+#endif /* AWS_CRYPTOSDK_MAKE_COMMON_DATA_STRUCTURES_H */
+
--- a/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/Makefile
+++ b/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/Makefile
@@ -1,0 +1,39 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+# this file except in compliance with the License. A copy of the License is
+# located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+###########
+# if Makefile.local exists, use it. This provides a way to override the defaults
+sinclude ../Makefile.local
+#otherwise, use the default values
+include ../Makefile.local_default
+
+## At the moment this is not used in the Makefile.common of encryption-sdk
+UNWINDSET +=
+
+CBMCFLAGS +=
+
+ENTRY = aws_cryptosdk_serialize_frame_harness
+
+DEPENDENCIES +=	$(SRCDIR)/c-common-helper-src/make_common_data_structures.c
+DEPENDENCIES +=	$(SRCDIR)/c-common-helper-src/proof_allocators.c
+DEPENDENCIES +=	$(SRCDIR)/c-common-helper-src/utils.c
+DEPENDENCIES +=	$(SRCDIR)/c-common-src/byte_buf.c
+DEPENDENCIES +=	$(SRCDIR)/c-common-src/common.c
+DEPENDENCIES +=	$(SRCDIR)/c-common-src/error.c
+DEPENDENCIES +=	$(SRCDIR)/c-enc-sdk-src/cipher.c
+DEPENDENCIES +=	$(SRCDIR)/c-enc-sdk-src/framefmt.c
+DEPENDENCIES += $(SRCDIR)/helper-src/make_common_data_structures.c
+
+###########
+
+include ../Makefile.common
--- a/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/Makefile
+++ b/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/Makefile
@@ -24,14 +24,14 @@ CBMCFLAGS +=
 
 ENTRY = aws_cryptosdk_serialize_frame_harness
 
-DEPENDENCIES +=	$(SRCDIR)/c-common-helper-src/make_common_data_structures.c
-DEPENDENCIES +=	$(SRCDIR)/c-common-helper-src/proof_allocators.c
-DEPENDENCIES +=	$(SRCDIR)/c-common-helper-src/utils.c
-DEPENDENCIES +=	$(SRCDIR)/c-common-src/byte_buf.c
-DEPENDENCIES +=	$(SRCDIR)/c-common-src/common.c
-DEPENDENCIES +=	$(SRCDIR)/c-common-src/error.c
-DEPENDENCIES +=	$(SRCDIR)/c-enc-sdk-src/cipher.c
-DEPENDENCIES +=	$(SRCDIR)/c-enc-sdk-src/framefmt.c
+DEPENDENCIES += $(SRCDIR)/c-common-helper-src/make_common_data_structures.c
+DEPENDENCIES += $(SRCDIR)/c-common-helper-src/proof_allocators.c
+DEPENDENCIES += $(SRCDIR)/c-common-helper-src/utils.c
+DEPENDENCIES += $(SRCDIR)/c-common-src/byte_buf.c
+DEPENDENCIES += $(SRCDIR)/c-common-src/common.c
+DEPENDENCIES += $(SRCDIR)/c-common-src/error.c
+DEPENDENCIES += $(SRCDIR)/c-enc-sdk-src/cipher.c
+DEPENDENCIES += $(SRCDIR)/c-enc-sdk-src/framefmt.c
 DEPENDENCIES += $(SRCDIR)/helper-src/make_common_data_structures.c
 
 ###########
--- a/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
+++ b/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
@@ -23,7 +23,7 @@ void aws_cryptosdk_serialize_frame_harness() {
     struct aws_cryptosdk_frame frame;  // Preconditions assume not null
     size_t ciphertext_size;
     size_t plaintext_size;
-    struct aws_byte_buf ciphertext_buf;
+    struct aws_byte_buf ciphertext_buf;  // Preconditions assume not null
     struct aws_cryptosdk_alg_properties *alg_props;
 
     /* Assumptions about the function input (based on the
@@ -37,16 +37,21 @@ void aws_cryptosdk_serialize_frame_harness() {
     __CPROVER_assume(aws_cryptosdk_frame_has_valid_type(&frame));
 
     /* Save the old state of the ciphertext buffer */
-    uint8_t *old_ciphertext_buffer   = ciphertext_buf.buffer;
-    size_t old_ciphertext_buffer_len = ciphertext_buf.len;
+    uint8_t *old_ciphertext_buf_ptr = ciphertext_buf.buffer;
+    size_t old_ciphertext_buf_len   = ciphertext_buf.len;
+    struct store_byte_from_buffer old_byte_from_ciphertext_buf;
+    save_byte_from_array(ciphertext_buf.buffer, ciphertext_buf.len, &old_byte_from_ciphertext_buf);
 
     int rval = aws_cryptosdk_serialize_frame(&frame, &ciphertext_size, plaintext_size, &ciphertext_buf, alg_props);
     if (rval == AWS_OP_SUCCESS) {
         assert(aws_cryptosdk_frame_is_valid(&frame));
         assert(aws_cryptosdk_alg_properties_is_valid(alg_props));
-        assert(aws_cryptosdk_frame_serialized(&frame, alg_props, plaintext_size));
-        assert(ciphertext_buf.buffer == old_ciphertext_buffer);
-        assert(ciphertext_buf.len == old_ciphertext_buffer_len + ciphertext_size);
+        assert(aws_cryptosdk_frame_is_valid_serialized(&frame, alg_props, plaintext_size));
+        assert(ciphertext_buf.buffer == old_ciphertext_buf_ptr);
+        assert(ciphertext_buf.len == old_ciphertext_buf_len + ciphertext_size);
+        if (old_ciphertext_buf_len > 0) {
+            assert_byte_from_buffer_matches((uint8_t *)ciphertext_buf.buffer, &old_byte_from_ciphertext_buf);
+        }
     } else {
         // Assert that the ciphertext buffer is zeroed in case of failure
         assert_all_zeroes(ciphertext_buf.buffer, ciphertext_buf.capacity);
--- a/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
+++ b/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <aws/common/byte_buf.h>
+#include <aws/cryptosdk/private/framefmt.h>
+#include <make_common_data_structures.h>
+#include <proof_helpers/make_common_data_structures.h>
+
+void aws_cryptosdk_serialize_frame_harness() {
+    /* data structure */
+    struct aws_cryptosdk_frame frame;  // Preconditions assume not null
+    size_t ciphertext_size;
+    size_t plaintext_size;
+    struct aws_byte_buf ciphertext_buf;
+    struct aws_cryptosdk_alg_properties *alg_props;
+
+    /* Assumptions about the function input (based on the
+     * preconditions) */
+    ensure_byte_buf_has_allocated_buffer_member(&ciphertext_buf);
+    __CPROVER_assume(aws_byte_buf_is_valid(&ciphertext_buf));
+
+    ensure_alg_properties_is_allocated(&alg_props);
+    __CPROVER_assume(aws_cryptosdk_alg_properties_is_valid(alg_props));
+
+    __CPROVER_assume(aws_cryptosdk_frame_has_valid_type(&frame));
+
+    /* Save the old state of the ciphertext buffer */
+    uint8_t *old_ciphertext_buffer   = ciphertext_buf.buffer;
+    size_t old_ciphertext_buffer_len = ciphertext_buf.len;
+
+    int rval = aws_cryptosdk_serialize_frame(&frame, &ciphertext_size, plaintext_size, &ciphertext_buf, alg_props);
+    if (rval == AWS_OP_SUCCESS) {
+        assert(aws_cryptosdk_frame_is_valid(&frame));
+        assert(aws_cryptosdk_alg_properties_is_valid(alg_props));
+        assert(aws_cryptosdk_frame_serialized(&frame, alg_props, plaintext_size));
+        assert(ciphertext_buf.buffer == old_ciphertext_buffer);
+        assert(ciphertext_buf.len == old_ciphertext_buffer_len + ciphertext_size);
+    } else {
+        // Assert that the ciphertext buffer is zeroed in case of failure
+        assert_all_zeroes(ciphertext_buf.buffer, ciphertext_buf.capacity);
+        assert(ciphertext_buf.len == 0);
+    }
+}
--- a/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/cbmc-batch.yaml
+++ b/.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/cbmc-batch.yaml
@@ -1,0 +1,4 @@
+jobos: ubuntu16
+cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwind;1;--unwinding-assertions;--object-bits;8"
+goto: aws_cryptosdk_serialize_frame_harness.goto
+expected: "SUCCESSFUL"
--- a/.cbmc-batch/source/make_common_data_structures.c
+++ b/.cbmc-batch/source/make_common_data_structures.c
@@ -44,3 +44,7 @@ void ensure_sig_ctx_has_allocated_members(struct aws_cryptosdk_sig_ctx *ctx) {
         evp_md_ctx_set0_evp_pkey(ctx->ctx, ctx->pkey);
     }
 }
+
+void ensure_alg_properties_is_allocated(struct aws_cryptosdk_alg_properties **alg_props) {
+    *alg_props = aws_cryptosdk_alg_props(nondet_int());
+}
--- a/include/aws/cryptosdk/private/framefmt.h
+++ b/include/aws/cryptosdk/private/framefmt.h
@@ -55,7 +55,7 @@ bool aws_cryptosdk_frame_is_valid(const struct aws_cryptosdk_frame *const frame)
  * the [alg_props] algorithm properties and in the
  * [plaintext_size]. It also checks that the two buffers are empty.
  */
-bool aws_cryptosdk_frame_serialized(
+bool aws_cryptosdk_frame_is_valid_serialized(
     const struct aws_cryptosdk_frame *frame,
     const struct aws_cryptosdk_alg_properties *alg_props,
     size_t plaintext_size);
--- a/include/aws/cryptosdk/private/framefmt.h
+++ b/include/aws/cryptosdk/private/framefmt.h
@@ -17,6 +17,7 @@
 #define AWS_CRYPTOSDK_PRIVATE_FRAMEFMT_H
 
 #include <aws/common/byte_buf.h>
+#include <aws/common/common.h>
 #include <aws/cryptosdk/cipher.h>
 #include <aws/cryptosdk/private/cipher.h>
 
@@ -32,6 +33,37 @@ struct aws_cryptosdk_frame {
     /* A cursor to space for the AEAD tag in the ciphertext buffer */
     struct aws_byte_buf authtag;
 };
+
+// MAX_FRAME_SIZE = 2^32 - 1
+#define MAX_FRAME_SIZE 0xFFFFFFFF
+// MAX_FRAMES = 2^32 - 1
+#define MAX_FRAMES 0xFFFFFFFF
+// MAX_UNFRAMED_PLAINTEXT_SIZE = 2^36 - 32
+#define MAX_UNFRAMED_PLAINTEXT_SIZE 0xFFFFFFFE0ull
+
+/**
+ * Checks whether a frame struct is valid. At the moment this means
+ * that it checks the validity of the byte buffers and the fact that
+ * they should have NULL allocators.
+ */
+bool aws_cryptosdk_frame_is_valid(const struct aws_cryptosdk_frame *const frame);
+
+/**
+ * Checks whether a frame struct is correctly serialized in relation
+ * to a specific algorithm.  More precisely, it checks that the
+ * buffers in [frame] have the same sizes as the sizes described in
+ * the [alg_props] algorithm properties and in the
+ * [plaintext_size]. It also checks that the two buffers are empty.
+ */
+bool aws_cryptosdk_frame_serialized(
+    const struct aws_cryptosdk_frame *frame,
+    const struct aws_cryptosdk_alg_properties *alg_props,
+    size_t plaintext_size);
+
+/**
+ * Checks whether a frame has a valid frame type.
+ */
+bool aws_cryptosdk_frame_has_valid_type(const struct aws_cryptosdk_frame *frame);
 
 /**
  * Performs frame-type-specific work prior to writing a frame; writes out all
--- a/source/cipher.c
+++ b/source/cipher.c
@@ -88,6 +88,15 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
     }
 }
 
+bool aws_cryptosdk_alg_properties_is_valid(const struct aws_cryptosdk_alg_properties *const alg_props) {
+    if (alg_props == NULL) {
+        return false;
+    }
+    enum aws_cryptosdk_alg_id id                             = alg_props->alg_id;
+    const struct aws_cryptosdk_alg_properties *std_alg_props = aws_cryptosdk_alg_props(id);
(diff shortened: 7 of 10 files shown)


source/framefmt.c
```diff
@@ -235,16 +252,23 @@ int aws_cryptosdk_serialize_frame(
...
+    
     // We assume that the max frame size is equal to the plaintext size. This
     // lets us avoid having to pass in a redundant argument, avoids needing to
     // take a branch in serde_framed, and does not impact the serialized
     // output.
     state.max_frame_size = plaintext_size;
     state.plaintext_size = plaintext_size;
     // Currently all supported algorithms have plaintext = ciphertext size
-    state.ciphertext_size = plaintext_size;
+    state.ciphertext_size = 0;
```
angelhof: I modified this so that the assertion [in line 45 of the harness](https://github.com/aws/aws-encryption-sdk-c/pull/378/files#diff-ac6b8365f754b91ff670c36d2435058dR45) passes. Is that assertion reasonable?

source/framefmt.c
```diff
@@ -235,16 +252,23 @@ int aws_cryptosdk_serialize_frame(
     size_t plaintext_size,
     struct aws_byte_buf *ciphertext_buf,
     const struct aws_cryptosdk_alg_properties *alg_props) {
+    AWS_PRECONDITION(aws_cryptosdk_alg_properties_is_valid(alg_props));
```
angelhof: Is this a reasonable precondition? Could this ever be called with an invalid `alg_props` struct?

include/aws/cryptosdk/private/framefmt.h
```diff
@@ -33,6 +33,15 @@ struct aws_cryptosdk_frame {
     struct aws_byte_buf authtag;
 };
 
+#define MAX_PLAINTEXT_SIZE 65535
```
angelhof: Is there a better suited place to define an upper bound for plaintext size? Bounding the plaintext_size argument to serialize is important as mentioned further below.

source/framefmt.c
```diff
@@ -235,16 +252,23 @@ int aws_cryptosdk_serialize_frame(
...
     struct aws_cryptosdk_framestate state;
 
+    // The plaintext_size should be bound to prevent arithmetic
+    // overflows due to addition
+    if (plaintext_size > MAX_PLAINTEXT_SIZE) {
```
angelhof: The `plaintext_size` should be bounded because otherwise there could be an overflow in lines 170, 194, 206.

source/cipher.c
```diff
@@ -88,6 +88,62 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
+        alg_props1.signature_len == alg_props2.signature_len &&
+        alg_props1.alg_id == alg_props2.alg_id;
+}
+
+bool aws_cryptosdk_alg_properties_is_valid(const struct aws_cryptosdk_alg_properties *const alg_props) {
+    if (alg_props == NULL) {
+        return false;
+    }   
+    enum aws_cryptosdk_alg_id id = alg_props->alg_id;
+    struct aws_cryptosdk_alg_properties *std_alg_props = aws_cryptosdk_alg_props(id);
```
angelhof: Is it reasonable to expect that the `alg_props` structure will always be initialized as shown in `aws_cryptosdk_alg_props`? Can this function act as the specification of a valid `alg_props` structure as shown here?

source/framefmt.c
```diff
@@ -213,6 +214,22 @@ static inline int serde_nonframed(
     return AWS_ERROR_SUCCESS;
 }
 
+bool aws_cryptosdk_frame_is_valid(const struct aws_cryptosdk_frame *const frame) {
+    bool iv_valid = aws_byte_buf_is_valid(&frame->iv);
+    bool ciphertext_valid = aws_byte_buf_is_valid(&frame->ciphertext);
+    /* This happens when input plaintext size is 0 */
+    bool ciphertext_valid_zero = frame->ciphertext.len == 0 &&
```
angelhof: A byte buf with capacity 0 in c-common is only considered valid if its buffer pointer is `NULL`. However `serialize` (when given plaintext_size = 0) can initialize the `ciphertext` byte buffer in the returning frame to point to a valid location in `ciphertext_buf` but to have 0 capacity. IMO it should be required that `plaintext_size > 0` in order to prevent this behaviour. Alternatively, we could relax the validity (as shown here), but I think that there is no reason to allow `plaintext_size == 0`.

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/cbmc-batch.yaml
```diff
@@ -0,0 +1,4 @@
+jobos: ubuntu16
+cbmcflags: "--bounds-check;--div-by-zero-check;--float-overflow-check;--nan-check;--pointer-check;--pointer-overflow-check;--signed-overflow-check;--undefined-shift-check;--unsigned-overflow-check;--unwind;1;--unwinding-assertions;--object-bits;8"
+goto: aws_byte_cursor_advance_harness.goto
```
tautschnig: This name appears to be wrong, and thus CBMC fails to actually perform any verification.

angelhof: Thanks, I added the `cbmc-batch.yaml` rule in the Makefile.

angelhof:
@david-koenig could you review this PR and the comments that i have added?

There are also 3 failing csdk failing builds. Could it be the case that these builds use an older version of `aws-c-common`? I am asking because ESDK builds on my machine. Could you take a look?  (The failing vs builds seem to be unrelated to this PR as with the others)

include/aws/cryptosdk/private/framefmt.h
lucasmt: According to the Message Body Specification documents, I believe the upper bound should be 2^36 - 32, is that correct?

david-koenig:
In our CodeBuild automation on Windows we are pinning to the 1.7.36 version of the AWS SDK for C++:

https://github.com/aws/aws-encryption-sdk-c/blob/master/codebuild/common-windows.bat#L18

That version of the AWS SDK for C++ is auto-installing v0.3.0 of aws-c-common, v0.1.2 of aws-checksums, and v0.1.1 of aws-c-event-stream:

https://github.com/aws/aws-sdk-cpp/commit/1d0d397406526c03c732abb9444cef9c0fda728a

david-koenig:
Sorry, I slightly misstated things in the last comment. The Windows builds are failing before they ever get to your code for the same still unknown reason that is causing #376 and #377 Windows builds to break on the aws-c-event-stream library.

It is the Linux builds that are failing because you are trying to use aws-c-common functions later than the v0.3.0 version. The CodeBuild CI for Linux is also pinning to the 1.7.36 version of the AWS SDK for C++ and is therefore using all of the same versions of the dependencies I listed in the previous comment:

https://github.com/aws/aws-encryption-sdk-c/blob/master/codebuild/bin/install-aws-deps.sh#L65

danielsn:
@david-koenig given that the Windows CI is broken, would it make sense to remove the required tag from it so as to avoid blocking pull requests?

include/aws/cryptosdk/private/framefmt.h
mattsb42-aws: There are three limits:
* Max frame size is 2^32 - 1
* Max frames is 2^32 - 1
* Max plaintext size **for unframed messages** is 2^36 - 32

Given that this is header concerns framed messages, only the first two are relevant here.

As for where these constants should be defined, or if they might already be (they should be...), but this header is definitely the wrong place.

source/framefmt.c
mattsb42-aws: I'll defer to @david-koenig or @ttjsu-aws on what this setting actually does, but from my cursory check of this file, this is not an acceptable change. This seems to me like a major change to the behavior of this function.

As a general rule, don't blindly change code to make tests pass.

source/framefmt.c
mattsb42-aws: Plaintext of size 0 is explicitly allowed for processing by all the AWS Encryption SDK implementations and required in the feature spec.

source/cipher.c
```diff
@@ -88,6 +88,56 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
     }
 }
 
+static bool str_eq(const char *const str1, const char *const str2) {
```
mattsb42-aws: Is this comparison constant time?

source/cipher.c
```diff
@@ -88,6 +88,56 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
+    if ((str1 && !str2) || (!str1 && str2)) {
+        return false;
+    }
+    if (!str1 && !str2) {
+        return true;
+    }
+#ifdef CBMC
```
mattsb42-aws: Why is this overridden for CMBC?

source/cipher.c
```diff
@@ -88,6 +88,56 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
+    return true;
+#else
+    return strcmp(str1, str2) == 0;
+#endif
+}
+
+static bool alg_properties_equal(
+    const struct aws_cryptosdk_alg_properties alg_props1, const struct aws_cryptosdk_alg_properties alg_props2) {
+    return
+        /* Note: strcmp messes with CBMC so we should either define
```
mattsb42-aws: Howso? What is the effect? What causes it?

source/cipher.c
```diff
@@ -88,6 +88,56 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
+           this to be different for CBMC or bound the size of names */
+        str_eq(alg_props1.md_name, alg_props2.md_name) && str_eq(alg_props1.cipher_name, alg_props2.cipher_name) &&
```
mattsb42-aws: Readability:
For readability and consistency, please stick to one check per line.

source/framefmt.c
angelhof: I agree that this is a major change to the behavior of the function and that it should not just be merged. However, I thought it would be better to make a "suggested" change in order for the proposed assertion in line 45 of the harness to pass. In any case, is this assertion reasonable?

source/framefmt.c
```diff
@@ -213,6 +214,22 @@ static inline int serde_nonframed(
     return AWS_ERROR_SUCCESS;
 }
 
+bool aws_cryptosdk_frame_is_valid(const struct aws_cryptosdk_frame *const frame) {
```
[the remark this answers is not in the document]
mattsb42-aws: What is this meant to validate?

* A serialized frame format? (I'm guessing no given the content, but the name does not rule this out)
* An `aws_cryptosdk_frame` struct?
  * Before it has been filled?
  * After?

From reading the function and the way that it is used, I am guessing that the intention was to validate the contents of a filled `aws_crypto_sdk` struct? If so, then we should also take in an `aws_cryptosdk_alg_properties` and frame size and validate that:
1. The `iv` and `authtag` buffers contain the correct number of bytes.
1. The sequence number is in a valid range (0 < `sequence_number` < 2^32).
1. The `ciphertext` contains the correct number of bytes for the frame `type` value.
   * `== frame_size` bytes if `type` is not final
   * `<= frame_size` bytes if `type` is final

source/framefmt.c
mattsb42-aws: This seems reasonable, yes.

That said, for context, this is a defense in depth: no users should ever be calling this function directly.

source/framefmt.c
mattsb42-aws: Agreed, but see my comments about frame size limits.

source/framefmt.c
mattsb42-aws: Read deeper on the assertion and I think I see the problem you ran into. I'll comment over there.

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
```diff
@@ -0,0 +1,46 @@
...
+
+    /* Save the old state of the ciphertext buffer */
+    uint8_t *old_ciphertext_buffer   = ciphertext_buf.buffer;
+    size_t old_ciphertext_buffer_len = ciphertext_buf.len;
+
+    int rval = aws_cryptosdk_serialize_frame(&frame, &ciphertext_size, plaintext_size, &ciphertext_buf, &alg_props);
+    if (rval == AWS_OP_SUCCESS) {
+        assert(aws_cryptosdk_frame_is_valid(&frame));
+        assert(ciphertext_buf.buffer == old_ciphertext_buffer);
+        assert(ciphertext_buf.len == old_ciphertext_buffer_len + ciphertext_size);
```
mattsb42-aws: As you found, `aws_cryptosdk_serialize_frame` modifies the ciphertext size to track the total ciphertext created.
If you want to make this assertion, rather than changing the behavior of `aws_cryptosdk_serialize_frame` to set the ciphertext size to zero (and thus making this line a no-op),
you should be making a more correct assertion about what is actually supposed to be happening here.

One complication here is that `ciphertext_buf.len` is going to increase *not only* by the length of the encrypted ciphertext (equal to the plaintext size) but also by the lengths of the frame metadata.

Is there a helper function somewhere that will determine the size of a frame given the frame length, frame type, and algorithm suite?

angelhof: I didn't find any helper function that returns the size. We could probably derive one though based on the calls to `field_sized` and `field_be32/64` in `serde_framed`, `serde_last_frame`, and `serde_nonframed` (such as [this call](https://github.com/aws/aws-encryption-sdk-c/pull/378/files#diff-72d57ea9d90face9d6fb936b9bbbed84R169)). 

In general, before the modification that I did to the code, the returned `ciphertext_size` had the value `plaintext_size + ciphertext_buf.len - old_ciphertext_buf.len`.

lucasmt: In PR #379 for the deserialize function I have manually computed the size of the frame for each case. I can extract them into helper functions if those don't already exist.

include/aws/cryptosdk/private/framefmt.h
lucasmt: MAX_FRAME_SIZE seems to be defined in private/session.h, but maybe it would be good to have it in a more general header so it is accessible here as well.

From what I understand, this header concerns unframed messages as well, no? Since both serialize and deserialize call serde_nonframed depending on their arguments.

source/cipher.c
angelhof: I will remove these comparisons for now, and simplify the `aws_cryptosdk_alg_properties_is_valid` function, as these properties are not necessary for the `framefmt` harnesses. We will revisit and extend this validity function when we go over proof harnesses for `cipher.c`.

source/framefmt.c
```diff
@@ -271,6 +295,7 @@ int aws_cryptosdk_serialize_frame(
         return aws_raise_error(result);
```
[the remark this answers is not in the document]
angelhof: @ttjsu-aws Should we assert anything in case of error?

include/aws/cryptosdk/private/framefmt.h
angelhof: @ttjsu-aws Where should we define these bounds?

source/cipher.c
ttjsu-aws: `aws_cryptosdk_alg_prop` looks up and returns the algorithm properties for a particular algorithm id and is currently used to initialize `alg_props` within the SDK. Refer: [cipher.h#L102](https://github.com/aws/aws-encryption-sdk-c/blob/master/include/aws/cryptosdk/cipher.h#L102).  However,  one could also do a static initialization like we have done here: [cipher.c#L40](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/cipher.c#L40)

source/framefmt.c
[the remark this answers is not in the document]
ttjsu-aws: yes, this [framefmt.c#L270](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L270) does the zeroing.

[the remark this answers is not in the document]
angelhof: @ttjsu-aws `secure_zero` zeros out the whole buffer and its length. Is this expected behavior or should we expect that only the part of the ciphertext buffer that was modified during this function call should be zeroed out?

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
angelhof: @ttjsu-aws Is this assertion reasonable? If not, then what does the returned `ciphertext_size` represent?

source/framefmt.c
ttjsu-aws: The `ciphertext_size` should be equal to the plaintext_size here as we do not include IV/authtag. I don't recommend changing this value to zero here.

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
ttjsu-aws: The ciphertext_size represents the amount of ciphertext to write to the frame and this is the value returned by the `aws_cryptosdk_serialize_frame` function.  `ciphertext_buf` is tracking the cursor for the frames initialized so far.  Given that, this assertion is not a reasonable one.

angelhof: @ttjsu-aws  Thanks for the response! However, there is something here that still confuses me. The returned ciphertext size is increased every time some space is reserved in the ciphertext buffer (see the macro definition `field_sized(state, bufptr, size)`). What is a reasonable assertion for the returned `ciphertext_size` value if the call to `serialize` succeeded?

source/framefmt.c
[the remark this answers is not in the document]
ttjsu-aws: Yes, it is the expected behavior that the ciphertext_buf is cleared out completely in the case of an error.

source/framefmt.c
mattsb42-aws: After discussing this with @ttjsu-aws offline, I am in agreement: this fundamentally changes the behavior of the function and makes the behavior no longer comply with the spec.

source/framefmt.c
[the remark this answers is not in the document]
angelhof: Thanks

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
ttjsu-aws: The returned ciphertext_size is increased by the size of the IV/authtag (if the frame has one). The IV/authtag size is not included in the ciphertext_size earlier.  Refer [framefmt.c#L168 ](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L168) and [framefmt.c#L170 ](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L170 for framed data and [framefmt.c#L203](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L203) and [framefmt.c#L206](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L206)

angelhof: @ttjsu-aws  Isn't the `plaintext_size` added again in [https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L205](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L205) for nonframed and in [https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L169](https://github.com/aws/aws-encryption-sdk-c/blob/master/source/framefmt.c#L169) for framed? In total ciphertext_size will contain the plaintext_size twice, once in the initialization and once after calling `field_sized`.

angelhof: If that is the expected behaviour, is the correct assertion for the returned ciphertext_size `ciphertext_size == new_ciphertext_buf_len - old_ciphertext_buf_len + plaintext_size`?

source/framefmt.c
bdonlan: After reviewing this, the `field_helper` and `field_sized` macros already add the total number of bytes needed to serialize the ciphertext to this field, so if this is _not_ set to zero, then the final `*ciphertext_size` is too large.

The main effect of this, as far as I can see, is that the output size estimate will be too large (we use this value at https://github.com/aws/aws-encryption-sdk-c/blob/e74d52d7c1595808171d79e6b6a6c2f029449724/source/session_encrypt.c#L270 where it's copied to the output size estimate). Note however that we will still succeed if the buffer size is correct, as we will only trigger a short buffer error if we did in fact run out of output buffer space.

Based on this analysis I believe this = 0 initialization is correct and corrects a minor bug, and that this exposes a gap in our unit tests.

source/framefmt.c
```diff
@@ -271,6 +300,7 @@ int aws_cryptosdk_serialize_frame(
...
     } else {
         *ciphertext_buf = state.u.buffer;
+        AWS_POSTCONDITION(aws_cryptosdk_frame_is_valid(frame));
```
bdonlan: If we're adding postconditions, I would suggest adding a postcondition that the ciphertext_buf is advanced by exactly `state.ciphertext_size` on success.

source/framefmt.c
david-koenig: I looked over the code in question with @bdonlan and I concur with his assessment that changing the initialization to 0 is correct. However along with this change we should do two other things:

(1) Get rid of the comment on the previous line which is now irrelevant.
(2) Include a unit test that tests this behavior and which would have failed with the improper initialization as it was done before.

source/framefmt.c
angelhof: Thanks for the suggestion! At the moment we don't have the infrastructure to reason about the initial (before the function call) value of a variable at the postconditions. However, I have added this assertion in the [harness](https://github.com/aws/aws-encryption-sdk-c/pull/378/files/e74d52d7c1595808171d79e6b6a6c2f029449724#diff-ac6b8365f754b91ff670c36d2435058dR44).

source/framefmt.c
angelhof: Thanks for the response :) I will remove the comment and add a unit test that exercises this behaviour.

david-koenig: I also think that the change to fix this bug and add the unit test could be separated out into a different code review rather than being mixed in with all this harness adding stuff.

angelhof: I will do it that

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
ttjsu-aws: I seem to have misunderstood the problem with the field_sized and field_header macros, I trust that Bryan and Dave's comments have resolved this  questions.

angelhof: Thanks, they have

source/framefmt.c
```diff
@@ -235,16 +254,26 @@ int aws_cryptosdk_serialize_frame(
...
     const struct aws_cryptosdk_alg_properties *alg_props) {
+    AWS_PRECONDITION(aws_cryptosdk_alg_properties_is_valid(alg_props));
     struct aws_cryptosdk_framestate state;
 
+    // The plaintext_size should be bound to prevent arithmetic
+    // overflows due to addition
+    if ((frame->type == FRAME_TYPE_SINGLE && plaintext_size > MAX_UNFRAMED_PLAINTEXT_SIZE) ||
+        (frame->type != FRAME_TYPE_SINGLE && plaintext_size > MAX_FRAME_SIZE)) {
+        // Clear the ciphertext buffer
+        aws_byte_buf_secure_zero(ciphertext_buf);
```
lucasmt: @angelhof Should the Makefile be modified to link `c-common-helper-src/utils.goto`? When I run this harness I get a warning that `aws_byte_buf_secure_zero` has no body.

angelhof: Nice catch, thanks!

angelhof:
@danielsn could you review this PR?

source/framefmt.c
```diff
@@ -213,6 +214,65 @@ static inline int serde_nonframed(
     return AWS_ERROR_SUCCESS;
 }
 
+bool aws_cryptosdk_frame_is_valid(const struct aws_cryptosdk_frame *const frame) {
+    if (frame == NULL) {
+        return false;
+    }
+
+    bool frame_type_seq_valid = aws_cryptosdk_frame_has_valid_type(frame);
```
bdonlan: This doesn't check the sequence number, so should not be `seq_valid`

angelhof: Thanks for the catch.

.cbmc-batch/include/make_common_data_structures.h
```diff
@@ -0,0 +1,26 @@
+/*
```
danielsn: Include guard.

.cbmc-batch/include/make_common_data_structures.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include <aws/common/common.h>
+#include <aws/cryptosdk/cipher.h>
+#include <aws/cryptosdk/private/framefmt.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/proof_allocators.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+/* This function creates a valid alg properties data structure by
```
danielsn: ```/**```

angelhof:
@mattsb42-aws I have addressed the changes. Could you review/approve the PR?

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -0,0 +1,20 @@
...
+ * or in the "license" file accompanying this file. This file is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <make_common_data_structures.h>
+
+void ensure_alg_properties_is_allocated(struct aws_cryptosdk_alg_properties **alg_props) {
+    *alg_props = aws_cryptosdk_alg_props(nondet_int());
```
danielsn: Is this the best way to do it? Do we get max nondet this way?

source/cipher.c
```diff
@@ -88,6 +88,15 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
     }
 }
 
+bool aws_cryptosdk_alg_properties_is_valid(const struct aws_cryptosdk_alg_properties *const alg_props) {
+    if (alg_props == NULL) {
+        return false;
+    }
+    enum aws_cryptosdk_alg_id id                             = alg_props->alg_id;
+    const struct aws_cryptosdk_alg_properties *std_alg_props = aws_cryptosdk_alg_props(id);
+    return std_alg_props == alg_props;
```
danielsn: Brackets here?

source/framefmt.c
```diff
@@ -213,6 +214,65 @@ static inline int serde_nonframed(
...
+        return false;
+    }
+
+    bool frame_type_valid = aws_cryptosdk_frame_has_valid_type(frame);
+
+    bool iv_byte_buf_valid  = aws_byte_buf_is_valid(&frame->iv);
+    bool iv_byte_buf_static = frame->iv.allocator == NULL;
+
+    bool authtag_byte_buf_valid  = aws_byte_buf_is_valid(&frame->authtag);
+    bool authtag_byte_buf_static = frame->authtag.allocator == NULL;
```
danielsn: I would write this as 
```bool authtag_byte_buf_static = (frame->authtag.allocator == NULL);```

File: .cbmc-batch/source/make_common_data_structures.c (hunk 20 above)
Comment (angelhof) [reply .cbmc-batch/source/make_common_data_structures.c]: I will add a comment explaining that

File: source/cipher.c (hunk 21 above)
Comment (angelhof) [reply source/cipher.c]: Thanks, I will add brackets.

File: source/framefmt.c (hunk 22 above)
Comment (angelhof) [reply source/framefmt.c]: Thanks, I will add brackets.

File: source/cipher.c [hunk 23]
```diff
@@ -88,6 +88,15 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
     }
 }
 
+bool aws_cryptosdk_alg_properties_is_valid(const struct aws_cryptosdk_alg_properties *const alg_props) {
```
danielsn: The comment on `ensure_is_allocated` also belongs here, to explain the check that we're doing that this is a singleton.

include/aws/cryptosdk/private/framefmt.h
```diff
@@ -33,6 +34,37 @@ struct aws_cryptosdk_frame {
     struct aws_byte_buf authtag;
 };
 
+// MAX_FRAME_SIZE = 2^32 - 1
```
danielsn: I assume these magic numbers from from the spec. If so, there should be a reference to the part of the spec doc that defines this.

.cbmc-batch/include/make_common_data_structures.h
```diff
@@ -1,3 +1,6 @@
+#ifndef AWS_CRYPTOSDK_MAKE_COMMON_DATA_STRUCTURES_H
```
danielsn: Internally in CBMC code, we can use `pragma once`

.cbmc-batch/include/make_common_data_structures.h
```diff
@@ -25,3 +29,16 @@ void ensure_md_context_has_allocated_members(struct aws_cryptosdk_md_context *ct
...
+
+/**
+ * This function creates a valid alg properties data structure by
+ * calling the initializer in cipher.h. This is the only valid way to
+ * construct an alg_props structure, and that is why
+ * [aws_cryptosdk_alg_properties_is_valid] checks whether the pointer
+ * of alg_props is the same as the one returned after calling the
+ * [aws_cryptosdk_alg_props] initializer.
+ */
+void ensure_alg_properties_is_allocated(struct aws_cryptosdk_alg_properties **alg_props);
```
danielsn: I'm a little confused by this comment. What does it mean for the pointer to be the same? Looking below, it looks like the alg_props returns a singleton for each algorithm type, and we are checking that this is so.  This should be clearly documented in the relevant ESDK header if it's an expected invariant of the system.

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/Makefile
```diff
@@ -0,0 +1,39 @@
...
+include ../Makefile.local_default
+
+## At the moment this is not used in the Makefile.common of encryption-sdk
+UNWINDSET +=
+
+CBMCFLAGS +=
+
+ENTRY = aws_cryptosdk_serialize_frame_harness
+
+DEPENDENCIES +=	$(SRCDIR)/c-common-helper-src/make_common_data_structures.c
```
danielsn: spacing

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
```diff
@@ -0,0 +1,55 @@
...
+#include <aws/cryptosdk/private/framefmt.h>
+#include <make_common_data_structures.h>
+#include <proof_helpers/make_common_data_structures.h>
+
+void aws_cryptosdk_serialize_frame_harness() {
+    /* data structure */
+    struct aws_cryptosdk_frame frame;  // Preconditions assume not null
+    size_t ciphertext_size;
+    size_t plaintext_size;
+    struct aws_byte_buf ciphertext_buf;
```
danielsn: Does this have a non-null precondition

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
```diff
@@ -0,0 +1,55 @@
...
+    ensure_byte_buf_has_allocated_buffer_member(&ciphertext_buf);
+    __CPROVER_assume(aws_byte_buf_is_valid(&ciphertext_buf));
+
+    ensure_alg_properties_is_allocated(&alg_props);
+    __CPROVER_assume(aws_cryptosdk_alg_properties_is_valid(alg_props));
+
+    __CPROVER_assume(aws_cryptosdk_frame_has_valid_type(&frame));
+
+    /* Save the old state of the ciphertext buffer */
+    uint8_t *old_ciphertext_buffer   = ciphertext_buf.buffer;
```
danielsn: This checks that the pointer didn't change, but not the existing bytes in the buffer.  We should probably check both things.

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -44,3 +44,7 @@ void ensure_sig_ctx_has_allocated_members(struct aws_cryptosdk_sig_ctx *ctx) {
         evp_md_ctx_set0_evp_pkey(ctx->ctx, ctx->pkey);
     }
 }
+
+void ensure_alg_properties_is_allocated(struct aws_cryptosdk_alg_properties **alg_props) {
+    *alg_props = aws_cryptosdk_alg_props(nondet_int());
```
danielsn: This is commented in the header, but is sufficiently unexpected to also be worth commenting here.

source/framefmt.c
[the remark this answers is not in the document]
danielsn: It looks like these are now being checked in the `aws_cryptosdk_frame_serialized` function.  Can someone on the service team confirm this now has the right meaning.

source/framefmt.c
```diff
@@ -213,6 +214,66 @@ static inline int serde_nonframed(
...
+        (frame->ciphertext.len == 0 && frame->ciphertext.buffer && frame->ciphertext.capacity == 0);
+    bool ciphertext_valid  = (ciphertext_byte_buf_valid || ciphertext_valid_zero);
+    bool ciphertext_static = (frame->ciphertext.allocator == NULL);
+
+    return (
+        frame_type_valid && iv_byte_buf_valid && iv_byte_buf_static && authtag_byte_buf_valid &&
+        authtag_byte_buf_static && ciphertext_valid && ciphertext_static);
+}
+
+bool aws_cryptosdk_frame_serialized(
```
danielsn: This function could have a better name.

source/framefmt.c
```diff
@@ -213,6 +214,66 @@ static inline int serde_nonframed(
...
+    const struct aws_cryptosdk_frame *frame,
+    const struct aws_cryptosdk_alg_properties *alg_props,
+    size_t plaintext_size) {
+    if (frame == NULL || alg_props == NULL) {
```
danielsn: Instead of checking NULL, could check `AWS_OBJECT_PTR_IS_READABLE`

source/framefmt.c
```diff
@@ -213,6 +214,66 @@ static inline int serde_nonframed(
     return AWS_ERROR_SUCCESS;
 }
 
+bool aws_cryptosdk_frame_is_valid(const struct aws_cryptosdk_frame *const frame) {
+    if (frame == NULL) {
```
danielsn: Instead of checking NULL, could check `AWS_OBJECT_PTR_IS_READABLE`

source/framefmt.c
```diff
@@ -213,6 +214,66 @@ static inline int serde_nonframed(
...
+    // Check that the ciphertext buffer has the correct size
+    bool ciphertext_size_valid = ((frame->type == FRAME_TYPE_SINGLE || frame->type == FRAME_TYPE_FRAME) &&
+                                  frame->ciphertext.capacity == plaintext_size) ||
+                                 (frame->type == FRAME_TYPE_FINAL && frame->ciphertext.capacity <= plaintext_size);
+
+    return (iv_size_valid && tag_size_valid && iv_empty && tag_empty);
+}
+
+bool aws_cryptosdk_frame_has_valid_type(const struct aws_cryptosdk_frame *frame) {
+    if (frame == NULL) {
```
danielsn: Instead of checking `NULL`, could check `AWS_OBJECT_PTR_IS_READABLE`

angelhof: Should I change that in all `frame == NULL` checks?

angelhof: Nevermind, I saw that you made the same comment on all of them

source/framefmt.c
angelhof: What about `aws_cryptosdk_frame_is_valid_serialized`?

.cbmc-batch/include/make_common_data_structures.h
angelhof: If I remember correctly this is the only allowed way to construct alg_props. Could someone from the ESDK team confirm?

.cbmc-batch/jobs/aws_cryptosdk_serialize_frame/aws_cryptosdk_serialize_frame_harness.c
angelhof: Yes it does. I will make a comment :)

include/aws/cryptosdk/private/framefmt.h
angelhof: That makes sense. However, I don't have a link to the spec. Could someone point to it?

tawdry-audrey:
Rebased, updated, and merged in https://github.com/aws/aws-encryption-sdk-c/pull/485.

PR #379, opened 2019-06-17
Added proof harness for aws_cryptosdk_deserialize_frame in framefmt.c

lucasmt:
*Description of changes:*

1. Added proof harness for `aws_cryptosdk_deserialize_frame` function in `framefmt.c`
2. Added check to `serde_nonframed` function in `framefmt.c` that the plaintext_size value read from the buffer is within bounds, in order to prevent arithmetic overflows
3. Changed `max_frame_size` argument in `aws_cryptosdk_deserialize_frame` from `uint64_t` to `uint32_t`

Depends on #378 and PR [#405](https://github.com/awslabs/aws-c-common/pull/405) of C-Common (for CBMC check in aws_hton64 of byte_order.h).

### DIFF
--- a/.cbmc-batch/source/make_common_data_structures.c
+++ b/.cbmc-batch/source/make_common_data_structures.c
@@ -15,11 +15,6 @@
 
 #include <make_common_data_structures.h>
 
-void ensure_alg_properties_has_allocated_names(struct aws_cryptosdk_alg_properties *const alg_props) {
-    size_t md_name_size;
-    alg_props->md_name = can_fail_malloc(md_name_size);
-    size_t cipher_name_size;
-    alg_props->cipher_name = can_fail_malloc(cipher_name_size);
-    size_t alg_name_size;
-    alg_props->alg_name = can_fail_malloc(alg_name_size);
+void ensure_alg_properties_is_allocated(struct aws_cryptosdk_alg_properties **alg_props) {
+    *alg_props = aws_cryptosdk_alg_props(nondet_int());
 }
--- a/.cbmc-batch/source/make_common_data_structures.c
+++ b/.cbmc-batch/source/make_common_data_structures.c
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+ * this file except in compliance with the License. A copy of the License is
+ * located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <make_common_data_structures.h>
+
+void ensure_alg_properties_has_allocated_names(struct aws_cryptosdk_alg_properties *const alg_props) {
+    size_t md_name_size;
+    alg_props->md_name = can_fail_malloc(md_name_size);
+    size_t cipher_name_size;
+    alg_props->cipher_name = can_fail_malloc(cipher_name_size);
+    size_t alg_name_size;
+    alg_props->alg_name = can_fail_malloc(alg_name_size);
+}
--- a/.cbmc-batch/source/openssl/objects_override.c
+++ b/.cbmc-batch/source/openssl/objects_override.c
@@ -22,14 +22,14 @@
  * numerical representation of an object. Return values: OBJ_txt2nid() returns a NID or NID_undef on error.
  */
 int OBJ_txt2nid(const char *s) {
-  // Currently these are the only values used in the ESDK
+    // Currently these are the only values used in the ESDK
     assert(!s || strcmp(s, "prime256v1") == 0 || strcmp(s, "secp384r1") == 0);
 
     if (!s) {
-      return NID_undef;
+        return NID_undef;
     } else if (strcmp(s, "prime256v1") == 0) {
         return NID_X9_62_prime256v1;
-    } else { // s is "secp384r1"
+    } else {  // s is "secp384r1"
         return NID_secp384r1;
     }
 }
--- a/.cbmc-batch/source/openssl/objects_override.c
+++ b/.cbmc-batch/source/openssl/objects_override.c
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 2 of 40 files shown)


source/framefmt.c
```diff
@@ -202,6 +203,12 @@ static inline int serde_nonframed(
     struct aws_cryptosdk_framestate *AWS_RESTRICT state, struct aws_cryptosdk_frame *AWS_RESTRICT frame) {
     field_sized(state, &frame->iv, state->alg_props->iv_len);
     field_be64(state, &state->plaintext_size);
+
+    // Check that plaintext_size is within bounds so we avoid an overflow later
+    if (state->plaintext_size >= MAX_PLAINTEXT_SIZE) {
```
lucasmt: This check prevents an arithmetic overflow in lines 212 or 213.

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -0,0 +1,25 @@
...
+ * or in the "license" file accompanying this file. This file is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <make_common_data_structures.h>
+
+void ensure_alg_properties_has_allocated_names(struct aws_cryptosdk_alg_properties *const alg_props) {
+    size_t md_name_size;
```
danielsn: these are all unconstrained? There are no max sizes the esdk enforces?

source/cipher.c
```diff
@@ -88,6 +88,56 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
+        return true;
+    }
+#ifdef CBMC
+    /* TODO: Bound the alg_props_string size to 128 characters (or any
+       reasonable max size).  This is only useful for CBMC to work,
+       not a real precondition. (We have to include that in the
+       validity check for alg_props (checking for that bound) so that
+       it is at least checked in debug_mode. */
+    /* For now just assert that they both point somewhere */
+    return true;
```
danielsn: Shoudn't this return nondet_bool()

source/cipher.c
```diff
@@ -88,6 +88,56 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
 
+static bool str_eq(const char *const str1, const char *const str2) {
+    if ((str1 && !str2) || (!str1 && str2)) {
+        return false;
+    }
+    if (!str1 && !str2) {
+        return true;
+    }
+#ifdef CBMC
+    /* TODO: Bound the alg_props_string size to 128 characters (or any
```
danielsn: assert that they're valid c-strings?

source/cipher.c
```diff
@@ -88,6 +88,56 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
+        alg_props1.content_key_len == alg_props2.content_key_len && alg_props1.iv_len == alg_props2.iv_len &&
+        alg_props1.tag_len == alg_props2.tag_len && alg_props1.signature_len == alg_props2.signature_len &&
+        alg_props1.alg_id == alg_props2.alg_id;
+}
+
+bool aws_cryptosdk_alg_properties_is_valid(const struct aws_cryptosdk_alg_properties *const alg_props) {
+    if (alg_props == NULL) {
+        return false;
+    }
+    enum aws_cryptosdk_alg_id id                             = alg_props->alg_id;
```
danielsn: IS this the clang-format they use?

source/framefmt.c
danielsn: AWS_ERROR_PRECONDITION once that is merged.

lucasmt: Does it make sense to have a precondition in the middle of a function?

lucasmt:
@ttjsu-aws @bdonlan I have updated this PR changing max_frame_size to a uint32_t, like we discussed yesterday.

.cbmc-batch/source/openssl/objects_override.c
```diff
@@ -0,0 +1,35 @@
...
+
+#include <proof_helpers/nondet.h>
+
+/*
+ * Description: OBJ_txt2nid() returns NID corresponding to text string <s>. s can be a long name, a short name or the
+ * numerical representation of an object. Return values: OBJ_txt2nid() returns a NID or NID_undef on error.
+ */
+int OBJ_txt2nid(const char *s) {
+    // Currently these are the only values used in the ESDK
+    assert(!s || strcmp(s, "prime256v1") == 0 || strcmp(s, "secp384r1") == 0);
```
danielsn: Doing this check twice might be slightly expensive.  Might be better to just add an `assert(0)` at the end of the if chain at the bottom.

.cbmc-batch/source/openssl/objects_override.c
```diff
@@ -0,0 +1,35 @@
...
+
+    if (!s) {
```
danielsn: Either `!s` is impossible, in which case this test doesn't make sense, or its a valid input with an error code, in which case it shouldn't be in the assert.

.cbmc-batch/source/openssl/objects_override.c
```diff
@@ -0,0 +1,35 @@
...
+        return NID_undef;
+    } else if (strcmp(s, "prime256v1") == 0) {
+        return NID_X9_62_prime256v1;
+    } else {  // s is "secp384r1"
+        return NID_secp384r1;
```
danielsn: As mentioned above, make this an `else if` and then we can avoid having to worry about the assert.

.cbmc-batch/source/proof_allocators.c
```diff
@@ -16,14 +16,16 @@
 #include "proof_allocators.h"
 #include <stdlib.h>
 
-//Use the same functions as the standard default allocator, but nondeterministically have malloc
-//return null.  This is needed because the CBMC model of malloc cannot fail, so we cannot
-//test the null case otherwise.
+// Use the same functions as the standard default allocator, but nondeterministically have malloc
```
danielsn: We can use the one from aws-c-common

include/aws/cryptosdk/cipher.h
```diff
@@ -101,6 +101,12 @@ struct aws_cryptosdk_alg_properties {
 AWS_CRYPTOSDK_API
 const struct aws_cryptosdk_alg_properties *aws_cryptosdk_alg_props(enum aws_cryptosdk_alg_id alg_id);
 
+/**
+ * Checks whether an aws_cryptosdk_alg_properties is valid and is supported by the SDK.
```
danielsn: We should note the runtime here, which I believe is O(1)

tawdry-audrey:
Rebased, updated, and merged in https://github.com/aws/aws-encryption-sdk-c/pull/485.

PR #381, opened 2019-07-10
Replace OpenSSL headers needed for the proofs with local copies

lucasmt:
*Description of changes:*

1. Created copies of the crypto.h, ec.h, ends.h, err.h, evp.h, openssl.h, ossl_typ.h, pem.h, rand.h and rsa.h OpenSSL headers inside the .cbmc-batch folder, with only the declarations and definitions that we need so far.
2. Changed Makefile.common to not add the OpenSSL path to the includes.

### DIFF
--- a/.cbmc-batch/include/openssl/asn1.h
+++ b/.cbmc-batch/include/openssl/asn1.h
@@ -1,0 +1,28 @@
+/*
+ * Changes to OpenSSL version 1.1.1. copyright 2019 Amazon.com, Inc. All Rights Reserved.
+ * Copyright 1995-2017 The OpenSSL Project Authors. All Rights Reserved.
+ *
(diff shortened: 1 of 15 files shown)


angelhof:
Is there a way to get the headers using git (similarly to the link-farms in the Makefile) so that there is no risk that they lag behind (and so that they don't need to be maintained)?

lucasmt:
@bdonlan @danielsn

.cbmc-batch/include/openssl/asn1.h
```diff
@@ -0,0 +1,7 @@
+#include <openssl/ossl_typ.h>
```
bdonlan: Some of these files are missing copyright attributions.

PR #385, opened 2019-07-10
Fix a bug where framefmt serialize returns a wrong ciphertext size

angelhof:
Fix the bug found in #378 and add a unit test to exercise it.


angelhof:
@david-koenig @ttjsu-aws could you review this PR?

tests/unit/t_framefmt.c
```diff
@@ -0,0 +1,61 @@
...
+    // Allocate and initialize the alg_properties
+    const struct aws_cryptosdk_alg_properties *alg_props = aws_cryptosdk_alg_props(id);
+    size_t ciphertext_size;
+
+    size_t old_ciphertext_buf_len = ciphertext_buf.len;
+
+    TEST_ASSERT_SUCCESS(
+        aws_cryptosdk_serialize_frame(&out_frame, &ciphertext_size, plaintext_size, &ciphertext_buf, alg_props));
+    // Assert that the returned size is correct
+    TEST_ASSERT(old_ciphertext_buf_len + ciphertext_size == ciphertext_buf.len);
```
david-koenig: If I understand correctly what is going on here, ``aws_cryptosdk_serialize_frame`` is supposed to take a buffer which might already have some bytes in use and then append the serialized frame to the end of it. Since you have only called ``aws_byte_buf_init`` on ``ciphertext_buf`` previously, the length will be initialized to zero. So in this case the sum of the left hand side of your assert is always just equal to ``ciphertext_size``. It would be a more meaningful test to verify that this works when there already is data in the buffer, which you could simulate by just setting ``ciphertext_buf.len`` to something nonzero before the call to ``aws_cryptosdk_serialize_frame``.

tests/unit/t_framefmt.c
```diff
@@ -0,0 +1,64 @@
...
+    // Alocate the frame
+    struct aws_cryptosdk_frame out_frame;
+    out_frame.type            = FRAME_TYPE_SINGLE;
+    out_frame.sequence_number = 1;
+
+    // Allocate the byte buffer
+    struct aws_byte_buf ciphertext_buf;
+    TEST_ASSERT_SUCCESS(aws_byte_buf_init(&ciphertext_buf, alloc, ciphertext_buf_capacity));
+
+    // Increase the length of the ciphertext to simulate having old data
```
david-koenig: Okay, I'm just nitpicking about terminology now, but tests are supposed to communicate what the code is supposed to do. If we have confusing comments it may mislead future developers. When you say "old data" it makes it sound like there is stale data in the buffer that's no longer relevant. That's not what's going on here. What would be in the buffer already is the earlier part of the ciphertext message, for example the header or the previous serialized frame.

Change ``old data`` in this comment to ``data already in the buffer``.

PR #392, opened 2019-07-24
Abstract model for OpenSSL ec.h module

lucasmt:
*Description of changes:*

1. Created file `ec_override.c` with abstractions for the `EC_GROUP` and `EC_KEY` structs.
2. Added helper functions for interacting with these data structures inside proofs.
3. Created stubs for OpenSSL functions that operate on these data structures.

### DIFF
--- a/.cbmc-batch/source/openssl/ec_override.c
+++ b/.cbmc-batch/source/openssl/ec_override.c
@@ -1,0 +1,456 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <limits.h>
+
+#include <bn_utils.h>
+#include <ec_utils.h>
+#include <openssl/ec.h>
+#include <openssl/objects.h>
+
+#include <proof_helpers/nondet.h>
+#include <proof_helpers/proof_allocators.h>
+
+/* Abstraction of the EC_GROUP struct */
+struct ec_group_st {
+    int curve_name;
+    point_conversion_form_t asn1_form;
+    BIGNUM *order;
+};
+
+/*
+ * Description: In order to construct a builtin curve use the function EC_GROUP_new_by_curve_name and provide the nid of
+ * the curve to be constructed. Return values: All EC_GROUP_new* functions return a pointer to the newly constructed
+ * group, or NULL on error.
+ */
+EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
+    assert(nid == NID_X9_62_prime256v1 || nid == NID_secp384r1);
+
+    EC_GROUP *group = can_fail_malloc(sizeof(EC_GROUP));
+
+    if (group) {
+        group->curve_name = nid;
+        group->asn1_form  = POINT_CONVERSION_UNCOMPRESSED;
+        group->order      = bignum_nondet_alloc();
+        __CPROVER_assume(bignum_is_valid(group->order));
+    }
+
+    return group;
+}
+
+/*
+ * Description: The functions EC_GROUP_set_point_conversion_form and EC_GROUP_get_point_conversion_form set and get the
+ * point_conversion_form for the curve respectively.
+ */
+void EC_GROUP_set_point_conversion_form(EC_GROUP *group, point_conversion_form_t form) {
+    assert(group);
+    group->asn1_form = form;
+}
+
+/*
+ * Return values: EC_GROUP_get0_order() returns an internal pointer to the group order.
+ */
+const BIGNUM *EC_GROUP_get0_order(const EC_GROUP *group) {
+    assert(ec_group_is_valid(group));
+    return group->order;
+}
+
+/*
+ * Description: EC_GROUP_free frees the memory associated with the EC_GROUP. If group is NULL nothing is done.
+ */
+void EC_GROUP_free(EC_GROUP *group) {
+    if (group) {
+        BN_free(group->order);
+        free(group);
+    }
+}
+
+/* Abstraction of the EC_KEY struct */
+struct ec_key_st {
+    int references;
+    EC_GROUP *group;
+    point_conversion_form_t conv_form;
+    BIGNUM *priv_key;
+    bool pub_key_is_set;  // We never have to return a public-key object, so just having this flag is enough
+};
+
+/*
+ * Description: A new EC_KEY with no associated curve can be constructed by calling EC_KEY_new(). The reference count
+ * for the newly created EC_KEY is initially set to 1. Return value: EC_KEY_new(), EC_KEY_new_by_curve_name() and
+ * EC_KEY_dup() return a pointer to the newly created EC_KEY object, or NULL on error.
+ */
+EC_KEY *EC_KEY_new() {
+    EC_KEY *key = can_fail_malloc(sizeof(EC_KEY));
+
+    if (key) {
+        key->references = 1;
+        key->group      = NULL;  // no associated curve
+        key->conv_form  = POINT_CONVERSION_UNCOMPRESSED;
+        // Can we assume that initially the keys are not set?
+        key->priv_key       = NULL;
+        key->pub_key_is_set = false;
+    }
+
+    return key;
+}
+
+/*
+ * Description: The function EC_KEY_get0_group() gets the EC_GROUP object for the key.
+ * Return values: EC_KEY_get0_group() returns the EC_GROUP associated with the EC_KEY.
+ */
+const EC_GROUP *EC_KEY_get0_group(const EC_KEY *key) {
+    assert(ec_key_is_valid(key));
+    return key->group;
+}
+
+/*
+ * Description: The function EC_KEY_set_group() sets the EC_GROUP object for the key.
+ * Return values: Returns 1 on success or 0 on error.
+ */
+int EC_KEY_set_group(EC_KEY *key, const EC_GROUP *group) {
+    assert(key);
+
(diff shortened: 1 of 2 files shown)


lucasmt:
@ttjsu-aws @bdonlan Could you review the OpenSSL models in `ec_override.c`?

source/cipher.c
```diff
@@ -88,6 +88,32 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
     }
 }
 
+static bool alg_properties_equal(
+    const struct aws_cryptosdk_alg_properties alg_props1, const struct aws_cryptosdk_alg_properties alg_props2) {
```
bdonlan: alg_properties structures should always be handled by reference, and should only point to the instances defined in cipher.h. It is expected that you should be able to confirm that two alg_props are equivalent by pointer equality.

source/cipher.c
```diff
@@ -88,6 +88,32 @@ static enum aws_cryptosdk_sha_version aws_cryptosdk_which_sha(enum aws_cryptosdk
...
+
+    /* Note: We are not checking whether the underlying alg_impl
+     * structs are equal. */
+    return alg_props1.data_key_len == alg_props2.data_key_len &&
+           alg_props1.content_key_len == alg_props2.content_key_len && alg_props1.iv_len == alg_props2.iv_len &&
+           alg_props1.tag_len == alg_props2.tag_len && alg_props1.signature_len == alg_props2.signature_len &&
+           alg_props1.alg_id == alg_props2.alg_id;
+}
+
+bool aws_cryptosdk_alg_properties_is_valid(const struct aws_cryptosdk_alg_properties *const alg_props) {
```
bdonlan: Nothing should be explicitly constructing alg_properties...?

.cbmc-batch/source/openssl/ec_override.c
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/** Decodes a ec public key from a octet string.
+ *  \param  key  a pointer to a EC_KEY object which should be used
+ *  \param  in   memory buffer with the encoded public key
+ *  \param  len  length of the encoded public key
+ *  \return EC_KEY object with decoded public key or NULL if an error
+ *          occurred.
+ */
+EC_KEY *o2i_ECPublicKey(EC_KEY **key, const unsigned char **in, long len) {
```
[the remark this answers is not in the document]
lucasmt: Could you point me to the man page where this behavior is documented? As far as I could tell from the OpenSSL implementation `*in` seems to always be advanced by `len`.

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+#include <make_common_data_structures.h>
+
+void ensure_alg_properties_has_allocated_names(struct aws_cryptosdk_alg_properties *const alg_props) {
+    size_t md_name_size;
+    alg_props->md_name = can_fail_malloc(md_name_size);
+    size_t cipher_name_size;
+    alg_props->cipher_name = can_fail_malloc(cipher_name_size);
```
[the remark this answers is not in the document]
lucasmt: This function was added in PR #378 and should be handled there. @angelhof can you take a look at these comments?

.cbmc-batch/source/openssl/ec_override.c
```diff
@@ -0,0 +1,246 @@
...
+ *  \param  key  the EC_KEY object with the public key
+ *  \param  out  the buffer for the result (if NULL the function returns number
+ *               of bytes needed).
+ *  \return 1 on success and 0 if an error occurred
+ */
+int i2o_ECPublicKey(const EC_KEY *key, unsigned char **out) {
+    assert(out);           // Assuming that it's always called in ESDK with non-NULL out
+    assert(*out == NULL);  // Assuming that it's always called with NULL *out, so buffer needs to be allocated
+
+    if (!key) return 0;
```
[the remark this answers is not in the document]
lucasmt: Could you point me to the documentation for this function as well? I couldn't find it for either `i2o_ECPublicKey` or `o2i_ECPublicKey`.

.cbmc-batch/source/openssl/ec_override.c
```diff
@@ -0,0 +1,246 @@
...
+
+    size_t buf_len;
```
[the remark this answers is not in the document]
lucasmt: Good catch, I'll correct that.

source/cipher.c
lucasmt: This function was added in PR #378 and should be handled there. @angelhof can you take a look at these comments?

source/cipher.c
lucasmt: This function was added in PR #378 and should be handled there. @angelhof can you take a look at these comments?

source/cipher.c
angelhof: Thanks for the comment @bdonlan . So I guess that is_valid should be something like that:
```
bool aws_cryptosdk_alg_properties_is_valid(const struct aws_cryptosdk_alg_properties *const alg_props) {
    if (alg_props == NULL) {
        return false;
    }
    enum aws_cryptosdk_alg_id id                             = alg_props->alg_id;
    const struct aws_cryptosdk_alg_properties *std_alg_props = aws_cryptosdk_alg_props(id);
    return std_alg_props == alg_props;
}
```

.cbmc-batch/source/make_common_data_structures.c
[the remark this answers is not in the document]
angelhof: I addressed this in https://github.com/danielsn/aws-encryption-sdk-c/pull/2.

source/cipher.c
angelhof: I addressed this in https://github.com/danielsn/aws-encryption-sdk-c/pull/2.

source/cipher.c
angelhof: I addressed this in https://github.com/danielsn/aws-encryption-sdk-c/pull/2.

lucasmt:
@bdonlan @danielsn

.cbmc-batch/source/openssl/ec_override.c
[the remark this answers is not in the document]
bdonlan: o2i_ECPublicKey itself does not appear to be well-documented, but o2i and d2i functions follow a consistent pattern, where the input pointer is advanced to follow the parsed data (see https://www.openssl.org/docs/man1.1.1/man3/d2i_ASIdentifiers.html for most of the functions with this pattern). If there is additional garbage after the public key, this should leave `*in` pointing to the garbage.

.cbmc-batch/source/openssl/ec_override.c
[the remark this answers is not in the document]
bdonlan: Can't see the original context of this due to the force-push, but if this is one of the `i2o/o2i_ECPublicKey` functions, the semantics of these functions appear to be undocumented (if they're documented, I can't find a mention in the manpages beyond simply a listing of the prototype), but they follow the naming convention of the d2i/i2d functions, which return a negative value on failure.

.cbmc-batch/source/openssl/ec_override.c
```diff
@@ -0,0 +1,443 @@
...
+EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
+    assert(nid == NID_X9_62_prime256v1 || nid == NID_secp384r1);
+
+    EC_GROUP *group = can_fail_malloc(sizeof(EC_GROUP));
+
+    if (group) {
+        group->curve_name = nid;
+        group->asn1_form  = POINT_CONVERSION_UNCOMPRESSED;
+	group->order = bignum_nondet_alloc();
+	__CPROVER_assume(bignum_is_valid(group->order)); // Can we assume order is always initialized?
```
bdonlan: I believe this is a safe assumption. If this fails the function should clean up and return null, so this failure case would be indistinguishable from a null return at line 41.

.cbmc-batch/source/openssl/ec_override.c
```diff
@@ -0,0 +1,443 @@
...
+ */
+EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
+    assert(nid == NID_X9_62_prime256v1 || nid == NID_secp384r1);
+
+    EC_GROUP *group = can_fail_malloc(sizeof(EC_GROUP));
+
+    if (group) {
+        group->curve_name = nid;
+        group->asn1_form  = POINT_CONVERSION_UNCOMPRESSED;
+	group->order = bignum_nondet_alloc();
```
bdonlan: Inconsistent indentation throughout.

.cbmc-batch/source/openssl/ec_override.c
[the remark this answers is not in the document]
bdonlan: https://github.com/openssl/openssl/blob/bbda79976b5c5095c5e6557311c86c623ba335f1/include/openssl/ec.h#L1118
https://github.com/openssl/openssl/blob/f585cefc046409d04f26b73307e677625be26e5a/crypto/ec/ec_asn1.c#L1105

Never mind, it is in fact zero. Oops.

.cbmc-batch/source/openssl/ec_override.c
```diff
@@ -0,0 +1,443 @@
...
+
+/*
+ * Description: ECDSA_SIG_get0() returns internal pointers the r and s values contained in sig and stores them in *pr and *ps, respectively. The pointer pr or ps can be NULL, in which case the corresponding value is not returned.
+ */
+void ECDSA_SIG_get0(const ECDSA_SIG *sig, const BIGNUM **pr, const BIGNUM **ps) {
+  assert(ecdsa_sig_is_valid(sig));
+  assert(pr);
+  assert(ps);
+
+  *pr = sig->r;
```
bdonlan: Should we check pr/ps for null and skip the assignment as documented?

lucasmt: We can do that if you think it's better, but as long as ESDK does not call it with NULL pointers the asserts should be enough.

.cbmc-batch/source/openssl/ec_override.c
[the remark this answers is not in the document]
lucasmt: Does the modification I made below to leave `*in` somewhere in between `*in` and `*in + len` look good?

.cbmc-batch/source/openssl/ec_override.c
lucasmt: Should be fixed now.

PR #403, opened 2019-08-02
Aws cryptosdk_enc_ctx (clear/init/clean_up) proofs

danielsn:

### DIFF
--- a/.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_clean_up-NULL/Makefile
+++ b/.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_clean_up-NULL/Makefile
@@ -1,0 +1,46 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+# this file except in compliance with the License. A copy of the License is
+# located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+#########
+# if Makefile.local exists, use it. This provides a way to override the defaults 
+sinclude ../Makefile.local
+#otherwise, use the default values
+include ../Makefile.local_default
+
+MAX_TABLE_SIZE ?= 4
+
+DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE)
+
+#A table has 10 words for the struct, plus 3 words for each entry
+TABLE_SIZE_IN_WORDS=$(shell echo $$(($$((3 * $(MAX_TABLE_SIZE))) + 10)))
+
+UNWINDSET += aws_hash_table_clear.0:$(shell echo $$((1 + $(MAX_TABLE_SIZE))))
+UNWINDSET += memset_override_0_impl.0:$(shell echo $$((1 + $(TABLE_SIZE_IN_WORDS))))
+
+CBMCFLAGS +=
+
+ENTRY = aws_cryptosdk_enc_ctx_clean_up_null_harness
+
+DEPENDENCIES += $(SRCDIR)/c-common-helper-src/make_common_data_structures.c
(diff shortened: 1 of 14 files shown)


danielsn:
This will not work in CI until dependecies are upgraded to aws-c-common later than Jul 29, 2019 a096d39540c497a58503ec6d6ba2ceed070e56df

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_clean_up-NULL/Makefile
```diff
@@ -0,0 +1,46 @@
...
+# if Makefile.local exists, use it. This provides a way to override the defaults 
+sinclude ../Makefile.local
+#otherwise, use the default values
+include ../Makefile.local_default
+
+MAX_TABLE_SIZE ?= 4
+
+DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE)
+
+#A table has 10 words for the struct, plus 3 words for each entry
```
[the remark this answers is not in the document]
danielsn: If this changes, then our loops bounds will be incorrect, which will be caught by the loop bound check.

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_clean_up-NULL/aws_cryptosdk_enc_ctx_clean_up_null_harness.c
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
karkhaz: 2019

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_clear/aws_cryptosdk_enc_ctx_clear_harness.c
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
karkhaz: 2019

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_clear/aws_cryptosdk_enc_ctx_clear_harness.c
```diff
@@ -0,0 +1,35 @@
...
+#include <proof_helpers/utils.h>
+
+void aws_cryptosdk_enc_ctx_clear_harness() {
+    struct aws_hash_table map;
+    ensure_allocated_hash_table(&map, MAX_TABLE_SIZE);
+    __CPROVER_assume(aws_hash_table_is_valid(&map));
+    ensure_hash_table_has_valid_destroy_functions(&map);
+
+    size_t empty_slot_idx;
+    __CPROVER_assume(aws_hash_table_has_an_empty_slot(&map, &empty_slot_idx));
```
karkhaz: It's not clear to me why you need to assume this for the proof to work

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_init/aws_cryptosdk_enc_ctx_init_harness.c
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
karkhaz: 2019

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_clear/aws_cryptosdk_enc_ctx_clear_harness.c
danielsn: In general, this is an important property of the hash-table, that ensures that certain hash-table functions terminate: they continue until they find the first empty slot, so if there is no empty slot, they never terminate.  This is a property that is ensured by the hash-table contract, so its safe to assume.  What I'd really like would be a way to ensure this in the ```aws_hash_table_is_valid()``` function, but I don't see an easy way that doesn't involve an expensive loop.

danielsn: That said, I think we're not calling any of the dangerous hash functions, so this may be an unneeded assumption.

PR #404, opened 2019-08-05
Enable clang-format on the CBMC-batch files

danielsn:
*Description of changes:*
cbmc proofs should also be clang-formatted.  

### DIFF
--- a/reformat.sh
+++ b/reformat.sh
@@ -15,4 +15,4 @@
 
 set -euxo pipefail
 
-find {.,aws-encryption-sdk-cpp}/{include,source,tests} examples -name '*.h' -or -name '*.c' -or -name '*.cpp' | xargs clang-format -i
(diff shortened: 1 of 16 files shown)


karlw00t:
Can we add a hook or CI that checks that this is done all the time? These point in time reformats tend to make git blame harder and if we are going to take the hit on a reformatting commit, we should make sure it happens on everything going forward.

reformat.sh
```diff
@@ -15,4 +15,4 @@
 
 set -euxo pipefail
 
-find {.,aws-encryption-sdk-cpp}/{include,source,tests} examples -name '*.h' -or -name '*.c' -or -name '*.cpp' | xargs clang-format -i
+find {.,aws-encryption-sdk-cpp}/{include,source,tests} .cbmc-batch examples -name '*.h' -or -name '*.c' -or -name '*.cpp' | xargs clang-format -i
```
danielsn: This line makes ```reformat.sh``` check the ```cbmc-batch``` folder.  On aws-c-common, this is checked in CI - not sure if this is also true for ESDK.

danielsn: @karlw00t re your comment.  Do you know if reformat.sh runs on CI?

PR #407, opened 2019-08-05
Abstract model for OpenSSL err.h module

lucasmt:
*Description of changes:*

Created stubs for OpenSSL functions in the `err.h` module.

### DIFF
--- a/.cbmc-batch/source/openssl/err_override.c
+++ b/.cbmc-batch/source/openssl/err_override.c
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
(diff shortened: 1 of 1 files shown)


lucasmt:
@bdonlan @danielsn

.cbmc-batch/source/openssl/err_override.c
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+#include <openssl/err.h>
+
+void ERR_print_errors_fp(FILE *fp) {
```
bdonlan: Hmm... do we have some of these lying around? If so we should remove them, we don't want to pollute the user's stderr.

lucasmt: There is one, in `aws_cryptosdk_sig_sign_start`. Should I submit a PR removing it? In that case this PR is unnecessary.

Issue #421, opened 2019-08-12
If EC has no elements, update single header field with the value 0 indicating no elements


ttjsu-aws:
Fix: https://github.com/aws/aws-encryption-sdk-c/blob/master/source/enc_ctx.c#L78

robin-aws:
Double-check if this is a breaking change

robin-aws:
Not a breaking change - the current behaviour is inconsistent with the specification and the other implementations, and means that for an empty EC, the C ESDK will create a header that can't be read by the other implementations.

Also deserialization currently rejects the EC if the header field is 0, which also must be fixed.

PR #422, opened 2019-08-12
Aws cryptosdk enc ctx serialize CBMC proof and fixup

danielsn:
*Issues* 
https://github.com/aws/aws-encryption-sdk-c/issues/421

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #425, opened 2019-08-16
Fix for Issue-421

danielsn:
*Issue #, if available:*
https://github.com/aws/aws-encryption-sdk-c/issues/421

*Description of changes:*
Fix the issue https://github.com/aws/aws-encryption-sdk-c/issues/421

### DIFF
--- a/source/enc_ctx.c
+++ b/source/enc_ctx.c
@@ -17,6 +17,7 @@
 #include <aws/cryptosdk/private/enc_ctx.h>
 #include <aws/cryptosdk/private/utils.h>
 
+#include <aws/common/assert.h>
 #include <aws/common/byte_buf.h>
 #include <aws/common/common.h>
 #include <aws/common/hash_table.h>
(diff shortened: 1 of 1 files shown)


source/enc_ctx.c
```diff
@@ -64,29 +65,25 @@ int aws_cryptosdk_enc_ctx_size(size_t *size, const struct aws_hash_table *enc_ct
...
+    if (!aws_byte_buf_write_be16(output, (uint16_t)num_elems)) {
+        aws_byte_buf_clean_up(output);
         return aws_raise_error(AWS_ERROR_SHORT_BUFFER);
     }
 
-    if (length == 0) {
-        // Empty encryption context
-        return AWS_OP_SUCCESS;
-    }
+    if (num_elems == 0) return AWS_OP_SUCCESS;
```
ttjsu-aws: If the num_elems is zero we want to add a value 0 to the header indicating that there are no elements. Line 71-72 is initializing  `output` to 0. This needs to be added only if the num_elems is 0 though.

danielsn: Line 71-72 is writing whatever the ```num_elems``` happens to be, whether that is ```0``` or non-zero.  If its zero, we've written our 0 and are done; if it is non-zero, we also, need to output the pairs themselves, which is what the loop below does.

ttjsu-aws: sgtm. 
Could you run the format script on this? 
Our styling for the if statement is different from this one.

source/enc_ctx.c
```diff
@@ -16,6 +16,7 @@
 #include <aws/cryptosdk/private/enc_ctx.h>
 #include <aws/cryptosdk/private/utils.h>
 
+#include <aws/common/assert.h>
```
dougch: Tests are failng to find this file, origin?

danielsn:
#422 superedes this

PR #426, opened 2019-08-16
ran reformat

danielsn:
*Description of changes:*
Run ```./reformat.sh``` on the code, to bring it up to the latest ```clang-format``` standards.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #430, opened 2019-08-22
Issue #322; Refreshing the clang-format file and checking the version…

### LINKED ISSUE #322: Better cross-platform compatibility for style checking
Current clang-format style check only works if you run `reformat.sh` script on Linux, not on Mac. Investigate reason for incompatible formatting with clang-format on Mac and see if we can fix this by modifying `.clang-format` file or using different version of the software.

Also investigate installing clang-format on Windows and whether we can get the same formatting there.

Failing better compatibility with clang-format, consider using a different style checker.

dougch:
resolves #322
* Used clang-format 8 to generate a default config file
* Folded differences from the old .clang-format file into the new version
* Fail clang-format versions before 8, as they fail to parse the config file
* Tested on Windows with version 9.0.0 from https://llvm.org/builds/

**Note:** there are 4 test files this configuration will update, those changes are not included in this PR.

```
$ git diff --stat
 tests/unit/t_header.c                  | 38 ++++++++++++++++++++++++++------------
 tests/unit/t_materials.c               | 12 ++++++------
 tests/unit/t_raw_aes_keyring_encrypt.c | 12 ++++++------
 tests/unit/t_raw_rsa_keyring_encrypt.c | 12 ++++++------
 4 files changed, 44 insertions(+), 30 deletions(-)
```


dougch:
PR #426 contains the changes to the tests

.clang-format
```diff
@@ -1,30 +1,150 @@
...
-# Function arguments should either be on one line, or should be each on their own line
-BinPackParameters: false
-BinPackArguments: false
-# Indent nested preprocessor directives
-IndentPPDirectives: AfterHash
-# Short conditionals are allowed:
-# case xyz: blah; break;
-# if (foo) return;
-# while (blah) {}
+Language:        Cpp
```
danielsn: Should this be "C"?

.clang-format
```diff
@@ -1,30 +1,150 @@
...
----
-Language: Cpp
-PointerAlignment: Right
-DerivePointerAlignment: false
-# Function arguments should either be on one line, or should be each on their own line
-BinPackParameters: false
-BinPackArguments: false
-# Indent nested preprocessor directives
-IndentPPDirectives: AfterHash
-# Short conditionals are allowed:
```
danielsn: Does the new style allow or disallow these?

dougch: These are all preserved as-is; the directives are sorted.

.clang-format
dougch: It was `Cpp` in the previous version of the file.  Clang-Format [docs](https://releases.llvm.org/8.0.0/tools/clang/docs/ClangFormatStyleOptions.html) don't show `C` as a valid option.

PR #462, opened 2020-01-09
Move macOS builds from travis-ci to github

dougch:
*Description of changes:*
Create a Github Action to build/test macOS.  Remove the travis-ci config currently controlling the build and move the script used to `examples` as it isn't needed for building (but is still useful for standalone builds).

# Check any applicable:
- [x] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.


dougch:
Tested on my personal fork, success [here](https://github.com/dougch/aws-encryption-sdk-c/commit/e7e7063ec24872d937f4e2071f02b9a8e25efd7c/checks?check_suite_id=392889639)

.github/workflows/osx.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Checkout Dependencies
+        uses: actions/checkout@v2
+        with:
+          repository: "aws/aws-sdk-cpp"
+          ref: "1.7.231"
+          path: "aws-sdk-cpp"
+
+      - name: Build and install aws-sdk-cpp
+        run: |
+          mkdir -p build-aws-sdk-cpp ||true
```
mattsb42-aws:
```suggestion
          mkdir -p build-aws-sdk-cpp || true
```

.github/workflows/osx.yml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  pull_request:
+    branches:
+      - master
+
+
+jobs:
+  OSX:
+    runs-on: macOS-10.14
```
mattsb42-aws: We should parameterize this and test with at least `macos-latest` as well.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategy

.github/workflows/osx.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+      - name: Install Dependencies
```
mattsb42-aws: FYI, `name` isn't required for steps; if you don't supply a name, GA just uses the name of the published action or the command (as appropriate for the step). Personally I like that better for simple steps like this, but YMMV.

.github/workflows/osx.yml
dougch: added

.github/workflows/osx.yml
```diff
@@ -0,0 +1,47 @@
...
+
+
+jobs:
+  OSX:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [macOS-10.14, macos-latest]
+        ssl_ver: [openssl@1.1]
```
mattsb42-aws: Bytes are cheap. Can we just spell this out? ie: `openssl_version`

.github/workflows/osx.yml
```diff
@@ -0,0 +1,47 @@
...
+          cd build-aws-sdk-cpp
+          cmake -G Xcode -DTARGET_ARCH="APPLE" -DBUILD_SHARED_LIBS=ON -DBUILD_ONLY="kms" -DENABLE_UNITY_BUILD=ON ../aws-sdk-cpp
+          xcodebuild -target ALL_BUILD
+          xcodebuild -target install
+
+      - name: Build C-ESDK
+        run: |
+          mkdir build-aws-encryption-sdk-c || true
+          cd build-aws-encryption-sdk-c
+          cmake -G Xcode -DBUILD_SHARED_LIBS=ON -DOPENSSL_ROOT_DIR="/usr/local/opt/openssl@1.1" ../
```
mattsb42-aws: Presumably, this should be referencing the matrix value?
I think the right way to inject this would be via an environment variable.

ex:
```yaml
- name: Build C-ESDK
  env:
    OPENSSL_VERSION: ${{ matrix.openssl_version }}
  run: |
    ...
    cmake -G -DBUILD_SHARED_LIBS=ON -DOPENSSL_ROOT_DIR="/usr/local/opt/${OPENSSL_VERSION}" ../
    ...
```

There's also the `format` function, but that feels messy for a large, multi-line, value like this.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#format

dougch:
Most recent runtimes on my fork: travis-ci: 18min vs github: 12min

.github/workflows/osx.yml
dougch: yup, good call.

.github/workflows/osx.yml
dougch: added.

Issue #450, opened 2019-10-03
Code does not compile against current aws-c-common


rmharris157:
The current state of **aws-encryption-sdk-c** (b5cb473) does not compile against the head of **aws-c-common** (tested using clang 8 on FreeBSD 11 AMI) ...

```
[  8%] Building C object CMakeFiles/aws-encryption-sdk.dir/source/error.c.o

/home/ec2-user/aws-encryption-sdk-c/source/error.c:33:5: error: implicit declaration of function 'aws_load_error_strings' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
    aws_load_error_strings();
    ^
1 error generated.
```

the `aws_load_error_strings` function was removed from **aws-c-common** on commit  7639117 (between v0.4.4 and v0.4.5).

(I am also _roberhr@amazon.com_ if you have other questions.)

dougch:
Correct, the README could use an upper bound on the version of c-common. Also see Issue #431 which has a link to the related PR.  This could be reprioritized after some of the build issues are resolved.

rmharris157:
As per ragona@, additional context here: aws-c-io will not compile unless you have aws-c-common >= v0.4.7 (function aws_common_library_init) missing. This is incompatible with the requirement to compile aws-c-common above with required version before v0.4.5.

dougch:
@rmharris157 can you clarify if your use case depends on aws-encryption-sdk-cpp ?

> As per ragona@, additional context here: aws-c-io will not compile unless you have aws-c-common >= v0.4.7 (function aws_common_library_init) missing. This is incompatible with the requirement to compile aws-c-common above with required version before v0.4.5.

dougch:
While `current` is a moving target, pr/463 should unblock the heart of this request

PR #463, opened 2020-01-17
Bump aws-cpp-sdk dependency to version 1.7.231 for CodeBuild (non-Linux)

dougch:
*Issue #450 *

*Description of changes:*
Version bump of dependency for platforms that install aws-cpp-sdk on every build.  The linux custom docker images contain a pre-installed aws-sdk-cpp and will be covered with a separate PR.

Also included, a CMake failure without explicit definition of the shared_libs flag.

- [x] Windows
- [x] macOS

# Check any applicable:


ttjsu-aws:
looks good! Seems like there is a build failure for `esdkCubuntu1804OpenSSL111`. Once that passes we can merge.

PR #472, opened 2020-02-17
Use GitHub Actions to publish docs

dougch:
*Description of changes:*
Automate away the creation of API docs.  Previously, the version numbers were put into versions.h as part of the build process, in order to remove the need to for a build/test cycle, moved version numbers into env variables.

Manual steps on the mac were:
```
(install all build deps)
mkdir build
cd build
cmake -G Xcode -DBUILD_SHARED_LIBS=OFF -DBUILD_DOC=ON -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl\@1.1 ..
xcodebuild -target doc_doxygen
git checkout gh-pages
rm -rf ../html
mv html ../
cd ../html
git add .
git commit -am "Docs for version xxx"
git push origin gh-pages
(open PR)
```

# Check any applicable:
- [x] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -1,0 +1,37 @@
+name: Documentation Builder
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - CHANGELOG.md
+
+jobs:
+  Doxygen:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+          ref: gh-pages
+      - name: Setup
+        run: |
+          sudo apt update && sudo apt install -y doxygen graphviz
+          cp include/aws/cryptosdk/version.h.in include/aws/cryptosdk/version.h
+          source versions
+          SOURCE_DIR=$(pwd) doxygen doxygen/doxygen.config
+      - name: Update html folder
+        run: |
+          if [ ! -d "doxygen/html" ]; then exit 255; fi
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          rm -rf html
+          mv doxygen/html .
+      - name: Upload docs branch
+          git commit -am "GHA Doxygen"
+      - name: Push docs
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: gh-pages
--- a/versions
+++ b/versions
@@ -1,0 +1,8 @@
+# An attempt to centralize version numbers consumable with bash.
+CESDK_MAJOR=1
+CESDK_MINOR=1
+CESDK_PATCH=0
+AWS_SDK_CPP=1.7.231
+AWS_C_COMMON=0.4.15
+AWS_CHECKSUMS=0.1.5
+AWS_C_EVENT_STREAM=0.1.4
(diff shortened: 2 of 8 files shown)


versions
```diff
@@ -0,0 +1,8 @@
+# An attempt to centralize version numbers consumable with bash.
+MAJOR=1
```
seebees: Would it be better to add some s2n value to this?
To disambiguate?  Or it that not how people use these?

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,38 @@
...
+          cp include/aws/cryptosdk/version.h.in include/aws/cryptosdk/version.h
+          source versions
+          SOURCE_DIR=$(pwd) doxygen doxygen/doxygen.config
+      - name: Commit
+        run: |
+          if [ ! -d "doxygen/html" ]; then exit 255; fi
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git fetch origin
+          git checkout gh-pages
```
seebees: https://github.com/actions/checkout/blob/master/action.yml#L7
Dose this mean that you can move this to the action?

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,38 @@
...
+          rm -rf html
+          mv doxygen/html .
+          git add --all
+          git commit -m "GHA Doxygen"
```
seebees: Can/should you `git commit -a`?  That will combine the line above...

seebees: It may be that `git add html/*` would be a better way.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Push docs
+        uses: ad-m/github-push-action@master
```
seebees: Why master and not a version?
This means that _anything_ that lands in master we now run...

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - name: Setup
+        run: |
+          sudo apt update && sudo apt install -y doxygen graphviz
+          cp include/aws/cryptosdk/version.h.in include/aws/cryptosdk/version.h
+          source versions
+          SOURCE_DIR=$(pwd) doxygen doxygen/doxygen.config
+      - name: Commit
```
seebees: This name is overloaded.  I would break up the "move the files around" and the really commit it :)

versions
dougch: Nice trying me to use the wrong project name...this was carried over from Cmake, I've made it more specific and then imported these ENV into Cmake variables.

.github/workflows/doxygen.yml
dougch: trying again with this approach

.github/workflows/doxygen.yml
dougch: dOxygen creates a static website with a few different sub-dirs, I don't want to accidentally leave any behind.

.github/workflows/doxygen.yml
dougch: Maintainer isn't cutting regular releases, 0.5.0 is behind master.

.github/workflows/doxygen.yml
dougch: Fixed.

dougch:
Obsolete

PR #509, opened 2020-05-01
Adds a proof harness for aws_cryptosdk_compare_hash_elems_by_key_string

feliperodri:
*Issue #, if available:*
N/A.

*Description of changes:*
- Adds a proof harness for `aws_cryptosdk_compare_hash_elems_by_key_string`;
- Disables enum range check for CBMC proofs.
- Properly removes all TAG-* files using `Makefile.common`.

# Check any applicable:


.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/Makefile
```diff
@@ -0,0 +1,31 @@
...
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+include ../Makefile.string
+
+CBMC_UNWINDSET += --unwindset memcmp.0:$(shell echo $$((1 + $(MAX_STRING_LEN))))
```
danielsn, line 16: Is this necessary?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,48 @@
...
+
+void aws_cryptosdk_compare_hash_elems_by_key_string_harness() {
+    /* Non-deterministic inputs. */
+    struct aws_hash_element elem_a;
+    elem_a.key = nondet_bool() ? ensure_string_is_allocated_bounded_length(MAX_STRING_LEN) : NULL;
+    __CPROVER_assume(aws_string_is_valid(elem_a.key));
+
+    struct aws_hash_element elem_b;
+    elem_b.key = nondet_bool() ? ensure_string_is_allocated_bounded_length(MAX_STRING_LEN) : NULL;
+    __CPROVER_assume(aws_string_is_valid(elem_b.key));
```
danielsn, line 29: does this make the previous line redundant?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,48 @@
...
+
+    bool nondet_parameter = nondet_bool();
+
+    /* Operation under verification. */
+    if (aws_cryptosdk_compare_hash_elems_by_key_string(&elem_a, nondet_parameter ? &elem_b : &elem_a) ==
+        AWS_OP_SUCCESS) {
+        const struct aws_string *key_a = (const struct aws_string *)elem_a.key;
+        const struct aws_string *key_b = (const struct aws_string *)elem_b.key;
+        if (nondet_parameter && key_a != NULL && key_b != NULL) {
```
danielsn, line 38: shouldn't this work whatever value `nondet_parameter` had?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,48 @@
...
+    /* Non-deterministic inputs. */
+    struct aws_hash_element elem_a;
+    elem_a.key = nondet_bool() ? ensure_string_is_allocated_bounded_length(MAX_STRING_LEN) : NULL;
+    __CPROVER_assume(aws_string_is_valid(elem_a.key));
+
+    struct aws_hash_element elem_b;
+    elem_b.key = nondet_bool() ? ensure_string_is_allocated_bounded_length(MAX_STRING_LEN) : NULL;
+    __CPROVER_assume(aws_string_is_valid(elem_b.key));
+
+    bool nondet_parameter = nondet_bool();
```
danielsn, line 31: Better name for this please

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,48 @@
...
+#include <aws/cryptosdk/private/utils.h>
+#include <make_common_data_structures.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/utils.h>
+
+void aws_cryptosdk_compare_hash_elems_by_key_string_harness() {
+    /* Non-deterministic inputs. */
+    struct aws_hash_element elem_a;
+    elem_a.key = nondet_bool() ? ensure_string_is_allocated_bounded_length(MAX_STRING_LEN) : NULL;
+    __CPROVER_assume(aws_string_is_valid(elem_a.key));
```
danielsn, line 25: We don't need to assume t that the element is valid?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,48 @@
...
+
+    struct aws_hash_element elem_b;
+    elem_b.key = nondet_bool() ? ensure_string_is_allocated_bounded_length(MAX_STRING_LEN) : NULL;
+    __CPROVER_assume(aws_string_is_valid(elem_b.key));
+
+    bool nondet_parameter = nondet_bool();
+
+    /* Operation under verification. */
+    if (aws_cryptosdk_compare_hash_elems_by_key_string(&elem_a, nondet_parameter ? &elem_b : &elem_a) ==
```
danielsn, line 34: No preconditions to add here?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/Makefile
feliperodri: Yes, this function calls `aws_string_compare` from AWS C Common, which calls `memcmp`.

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: Yes, I fixed it.

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: Yes, but if `nondet_parameter == false`then the byte comparison should be against `elem_a` itself and not `elem_b` (which would be redundant to add, in my opinion).

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: I renamed it to `nondet_lhs`, is it better?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: I've updated this portion in the harness. Now, we properly assume two cases: the element is valid and the element is null. In the case the element is != NULL, it must be valid; otherwise, it'll fail the precondition on function `aws_string_compare`.

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: Yep, good catch! I've added a pre-condition to check for nullness.

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,63 @@
...
+    /* Non-deterministic inputs. */
+    struct aws_hash_element *elem_a = can_fail_malloc(sizeof(*elem_a));
+    if (elem_a != NULL) {
+        if (nondet_bool()) {
+            elem_a->key = ensure_string_is_allocated_bounded_length(MAX_STRING_LEN);
+            __CPROVER_assume(aws_string_is_valid(elem_a->key));
+        } else {
+            elem_a->key = NULL;
+        }
+    }
```
danielsn, line 31: This code is duplicated - might be worth factoring out into a function

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,63 @@
...
+    if (elem_b != NULL) {
+        if (nondet_bool()) {
+            elem_b->key = ensure_string_is_allocated_bounded_length(MAX_STRING_LEN);
+            __CPROVER_assume(aws_string_is_valid(elem_b->key));
+        } else {
+            elem_b->key = NULL;
+        }
+    }
+
+    bool nondet_lhs = nondet_bool();
```
danielsn, line 43: maybe call this `nondet_alias_elements`

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
```diff
@@ -0,0 +1,63 @@
...
+
+    /* Pre-conditions. */
+    __CPROVER_assume(elem_a != NULL);
+    __CPROVER_assume(elem_b != NULL);
+
+    /* Operation under verification. */
+    if (aws_cryptosdk_compare_hash_elems_by_key_string(elem_a, nondet_lhs ? elem_b : elem_a) == AWS_OP_SUCCESS) {
```
danielsn, line 50: it might be clearer to do the aliasing when you create `elem_b`, so the assumptions all go at the top.  Not sure tho.

source/utils.c
```diff
@@ -17,6 +17,7 @@
 #include <aws/cryptosdk/private/utils.h>
 
 int aws_cryptosdk_compare_hash_elems_by_key_string(const void *elem_a, const void *elem_b) {
+    AWS_PRECONDITION(elem_a != NULL && elem_b != NULL);
     const struct aws_hash_element *a = (const struct aws_hash_element *)elem_a;
     const struct aws_hash_element *b = (const struct aws_hash_element *)elem_b;
```
danielsn, line 22: Are there `_is_valid()` preconditions on `a` and `b`?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: I know it is repetitive here, but this code is specific as well. It creates a hash element that stores a bounded string object (we could create the same thing for different objects). To avoid duplicity, I wrote a function, but I added it to the harness itself. We could add to AWS C Common a function for that (since it's related to the `aws_hash` module). What do you think?

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: Done.

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/aws_cryptosdk_compare_hash_elems_by_key_string_harness.c
feliperodri: Done.

source/utils.c
feliperodri: Nope, they're `void*`. Also, we check for their validity on the `aws_string_compare` call.

.cbmc-batch/jobs/aws_cryptosdk_compare_hash_elems_by_key_string/Makefile
```diff
@@ -0,0 +1,31 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
danielsn, line 1: copyright year no longer needed.

feliperodri: Done.

Issue #514, opened 2020-05-07
aws-sdk-cpp failing to compile in windows CI


danielsn:
Output from codebuild: 

```

C:\build>mkdir build-aws-sdk-cpp
181 |  
182 | C:\build>cd build-aws-sdk-cpp
183 |  
184 | C:\build\build-aws-sdk-cpp>cmake -DBUILD_SHARED_LIBS=OFF -G "Visual Studio 14 2015 Win64" -DCMAKE_INSTALL_PREFIX=c:/deps -DCMAKE_BUILD_TYPE="Release" -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake -DBUILD_ONLY="kms" -DENABLE_UNITY_BUILD=ON ../aws-sdk-cpp   \|\| goto error
185 | -- Found Git: C:/Program Files/Git/cmd/git.exe (found version "2.22.0.windows.1")
186 | -- TARGET_ARCH not specified; inferring host OS to be platform compilation target
187 | -- Building AWS libraries as static objects
188 | -- Generating windows build config
189 | -- Building project version: 1.7.231
190 | -- Selecting Windows SDK version  to target Windows 10.0.14393.
191 | CMake Error at CMakeLists.txt:7 (project):
192 | Failed to run MSBuild command:
193 |  
194 | C:/Program Files (x86)/MSBuild/14.0/bin/MSBuild.exe
195 |  
196 | to get the value of VCTargetsPath:
197 |  
198 | Exit code 0xc0000142
199 |  
200


```

dougch:
CodeBuild is updating the underlying windows version- need to rebuild our custom imaged based on microsoft/windowsservercore:ltsc2016

dougch:
Rebuild of Docker images in progress- latest security patch is also required in addition to the newer ltsc2016.

dougch:
recent pr #512 windows jobs all passing.

PR #517, opened 2020-05-08
Windows dockerupdate2020

dougch:
*Issue #, if available:*
#514 

*Description of changes:*
Rebuild Windows docker containers with updated base version.  This is required to run on CodeBuild.

Images names are vs2017-20200508-16 and vs2015-20200508-16

# Check any applicable:


docker-images/windows_base/Dockerfile
```diff
@@ -11,7 +11,9 @@ ADD https://download.microsoft.com/download/6/A/A/6AA4EDFF-645B-48C5-81CC-ED5963
 RUN start /wait C:\vc_redist.x64.exe /quiet /norestart
 
 # Install chocolatey
-RUN @powershell -NoProfile -ExecutionPolicy unrestricted -Command "(iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'))) >$null 2>&1"
+# which needs TLS1.2 for powershell too be enabled. it's 2020.
```
danielsn, line 14: We should probably have a ticket to check if that ever gets fixed on their end, and turn it off on our end too.

dougch: More reading on this [here](https://blog.pauby.com/post/force-powershell-to-use-tls-1-2/). It is pointed out that technically, this has todo with .NET versions (time to upgrade).

dougch:
> Is there a way to set up a github action to automate this next time?

Yes, but the tricky bit is automated Windows Docker image creation, when we looked at this a while back it was not...trivial.

PR #522, opened 2020-05-12
Use recommended RSA padding mode

juneb:
*Issue #, if available:*

*Description of changes:*
Changing the example to use the recommend OAEP padding mode. We discourage the use of the PKCS#1 padding mode. It's included only for backwards compatibility.

# Check any applicable:


texastony (approved):
LGTM. It is reformatting and then switching the RSA padding scheme.

PR #551, opened 2020-06-11
fix: base64_encode error in examples/caching_cmm.cpp

### LINKED ISSUE #550: Error with aws_base_encode for caching cmm
Hi, we're having issues with encrypting data locally with an IAM role for KMS access using caching_cmm.cpp with the error being thrown up as

Unexpected error in aws_base64_encode: Unknown Error Code

Just to be certain, we do not think its an IAM / permissions issue we are able to run string.cpp from the examples.

robin-aws:
Fixes #550

The example code was invoking `aws_byte_buf_from_array` on the vector's backing array, but this initializer creates a `byte_buf` that is already full (i.e. `len == capacity`), so the encoding logic would think there was `0` capacity left to encode into. The correct initializer is `aws_byte_buf_from_empty_array` which sets `len` to `0` instead and overwrites the array contents when written to.

Cut a separate issue for the fact that the CI is not executing the examples: #552 

# Check any applicable:

### DIFF
--- a/examples/caching_cmm.cpp
+++ b/examples/caching_cmm.cpp
@@ -92,25 +92,24 @@ std::string decrypt(
     return std::string(buffer.begin(), buffer.end());
 }
 
-std::string base64_encode(struct aws_allocator *alloc, const std::vector<uint8_t> &vec) {
+std::string base64_encode(const std::vector<uint8_t> &vec) {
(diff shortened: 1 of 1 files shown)


robin-aws:
Stale, already fixed in another merged PR

PR #567, opened 2020-08-02
Improves coverage for the aws_cryptosdk_edk_list_init proof

tegansb:
Description of changes:
Addresses the low coverage of the aws_cryptosdk_edk_list_init proof.
Includes a README for the aws_cryptosdk_edk_list_init proof specifying expected coverage behavior.
Adds the aws_cryptosdk_empty_edk_list_is_valid function to check for valid empty edk lists directly. Improves coverage when compared to the aws_cryptosdk_edk_list_is_valid function for empty edk lists. 

Coverage for aws_cryptosdk_edk_list_init improved from .71 to .97.

@danielsn @feliperodri 

# Check any applicable:

### DIFF
--- a/.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
+++ b/.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
@@ -7,8 +7,10 @@ Some functions contain unreachable blocks of code:
 
 * `aws_array_list_is_valid`
 
-    * list is never NULL
+    * list is never NULL as already ensured by a precondition 
 
 * `aws_mul_u64_checked`
 
-    *  Multiplication never results in an overflow. 
+    * Multiplication never results in an overflow
+    * The two arguments for mutliplication are the initial_item_allocation and sizeof(struct aws_cryptosdk_edk)
+    * The initial_item_allocation is 4 as hardcoded by function aws_cryptosdk_edk_list_init
--- a/.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
+++ b/.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
@@ -1,10 +1,12 @@
-# Expected Coverage 
+# Memory safety proof for aws_cryptosdk_edk_list_init
 
-0.97 (62 lines out of 64 statically-reachable lines in 13 functions reached)
-0.97 (62 lines out of 64 statically-reachable lines in 13 statically-reachable functions)
+This proof harness attains 97% code coverage.  The following comments explain
+why the uncovered lines of code are unreachable code.
 
-## Expected Functions with Incomplete Coverage 
+Some functions contain unreachable blocks of code:
 
-(10/11)	aws_array_list_is_valid: list is never NULL
+* aws_array_list_is_valid
+    * list is never NULL
 
-(4/5) aws_mul_u64_checked: Multiplication never results in an overflow. 
+* aws_mul_u64_checked
+    *  Multiplication never results in an overflow. 
--- a/.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
+++ b/.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
@@ -5,8 +5,10 @@ why the uncovered lines of code are unreachable code.
 
 Some functions contain unreachable blocks of code:
 
-* aws_array_list_is_valid
+* `aws_array_list_is_valid`
+
     * list is never NULL
 
-* aws_mul_u64_checked
+* `aws_mul_u64_checked`
+
     *  Multiplication never results in an overflow. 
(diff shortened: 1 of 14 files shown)


.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
```diff
@@ -0,0 +1,10 @@
+# Expected Coverage 
+
+0.97 (62 lines out of 64 statically-reachable lines in 13 functions reached)
+0.97 (62 lines out of 64 statically-reachable lines in 13 statically-reachable functions)
+
+## Expected Functions with Incomplete Coverage 
+
+(10/11)	aws_array_list_is_valid: list is never NULL
+
+(4/5) aws_mul_u64_checked: Multiplication never results in an overflow.
```
feliperodri, line 10: Could you follow the same style we used for [MQTT proofs](https://github.com/aws/aws-iot-device-sdk-embedded-C/tree/v4_beta/cbmc/proofs/IotMqtt_PublishAsync)?

source/edk.c
```diff
@@ -101,3 +101,9 @@ bool aws_cryptosdk_edk_list_is_valid(const struct aws_array_list *edk_list) {
     return true;
 #endif /* AWS_DEEP_CHECKS == 1 */
 }
+
+bool aws_cryptosdk_empty_edk_list_is_valid(const struct aws_array_list *edk_list) {
+    AWS_PRECONDITION(edk_list->length == 0);  // ensure that the list is empty
+    return AWS_OBJECT_PTR_IS_READABLE(edk_list) && aws_array_list_is_valid(edk_list) &&
+           (edk_list->item_size == sizeof(struct aws_cryptosdk_edk));
+}
```
feliperodri, line 109: In this case, `AWS_OBJECT_PTR_IS_READABLE(edk_list)` will always be true. In fact, you need to add it as a precondition, so you'll avoid the possibility of a null-pointer dereference in `edk_list->length`. Do we consider `edk_list == NULL` as a valid empty edk_list? If not, I'd suggest:
```c
bool aws_cryptosdk_empty_edk_list_is_valid(const struct aws_array_list *edk_list) {
    AWS_PRECONDITION(AWS_OBJECT_PTR_IS_READABLE(edk_list));
    AWS_PRECONDITION(edk_list->length == 0);
    return aws_array_list_is_valid(edk_list) &&
           (edk_list->item_size == sizeof(struct aws_cryptosdk_edk));
}
```
We can also remove the comment, since the code is pretty clear.

.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/README.md
tegansb: Yes! I've gone ahead and made these changes everywhere.

source/edk.c
tegansb: You're right. We do not consider edk_list == NULL as a valid empty list, so I've made the changes you suggested.

.cbmc-batch/jobs/aws_cryptosdk_edk_list_init/aws_cryptosdk_edk_list_init_harness.c
```diff
@@ -23,6 +23,6 @@ void aws_cryptosdk_edk_list_init_harness() {
 
     int rval = aws_cryptosdk_edk_list_init(can_fail_allocator(), &edk_list);
     if (rval == AWS_OP_SUCCESS) {
-        assert(aws_cryptosdk_edk_list_is_valid(&edk_list));
+        aws_cryptosdk_empty_edk_list_is_valid(&edk_list);
```
danielsn, line 26: Why removing this assert? This should stay as an assert

tegansb: Right, thanks for catching that!

source/edk.c
```diff
@@ -101,3 +101,9 @@ bool aws_cryptosdk_edk_list_is_valid(const struct aws_array_list *edk_list) {
     return true;
 #endif /* AWS_DEEP_CHECKS == 1 */
 }
+
+bool aws_cryptosdk_empty_edk_list_is_valid(const struct aws_array_list *edk_list) {
+    AWS_PRECONDITION(AWS_OBJECT_PTR_IS_READABLE(edk_list));
+    AWS_PRECONDITION(edk_list->length == 0);
+    return aws_array_list_is_valid(edk_list) && (edk_list->item_size == sizeof(struct aws_cryptosdk_edk));
+}
```
feliperodri, line 109: @robin-aws @alex-chew is this reasonable?

alex-chew: This looks reasonable to me.

tegansb: Great thank you!

danielsn (approved):

Issue #414, opened 2019-08-07
Incorporate the decrypt test vector into CI


seebees:
Test vectors here:
https://github.com/awslabs/aws-encryption-sdk-test-vectors

robin-aws:
To clarify (thanks @ttjsu-aws), the entry point to handle the test vectors is implemented (https://github.com/aws/aws-encryption-sdk-c/blob/master/aws-encryption-sdk-cpp/tests/test_vectors/static_test_vectors.cpp) but only run manually - you currently have to download the test vectors yourself to run them.

PR #571, opened 2020-08-07
Add aws-encryption-sdk-test-vectors as a submodule

### LINKED ISSUE #225: git sub-module for aws-encryption-sdk-test-vectors package

robin-aws:
Fixes #225 

TODO: script to unzip the vectors and run test_vectors, which should then be added to CI (#414)

# Check any applicable:


robin-aws:
> I think we need to open a new PR so that this goes into `aws:spec-1.0-alignment` instead of `aws:master`.

Since this only affects testing/CI I think it's fine to go into master personally.

alex-chew:
I see, that makes sense. All good then.

PR #568, opened 2020-08-03
Adds documentation for aws_cryptosdk_md_* proof coverage

tegansb:
Description of changes:
Addresses the low coverage of the aws_cryptosdk_md_init and aws_cryptosdk_md_finish proofs.
Includes a README for both the aws_cryptosdk_md_init and aws_cryptosdk_md_finish proofs specifying expected coverage behavior.
Uses AWS_FATAL_PRECONDITION instead of abort(). 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #574, opened 2020-08-12
Improves coverage for the aws_cryptosdk_md_{abort, update} proofs 

tegansb:
Description of changes:
Addresses the low coverage of the aws_cryptosdk_md_update and aws_cryptosdk_md_abort proofs.
Includes a README for both the aws_cryptosdk_md_update and aws_cryptosdk_md_abort proofs specifying expected coverage behavior.
Added stubs for EVP_MD_CTX_free and evp_md_ctx_is_valid to improve coverage. These stubs should only be used when ctx->pkey is NULL and they assert this condition. These stubs should also be used for the function in PR #568, which I will close and then re-open based on this PR. 
Coverage for aws_cryptosdk_md_update is improved from .69 to 1.0. 
Coverage for aws_cryptosdk_md_abort  is improved from .62 to 1.0. 

@danielsn @feliperodri


.cbmc-batch/jobs/aws_cryptosdk_md_abort/Makefile
```diff
@@ -35,6 +36,11 @@ DEPENDENCIES += $(SRCDIR)/helper-src/openssl/bn_override.goto
 DEPENDENCIES += $(SRCDIR)/helper-src/openssl/ec_override.goto
 DEPENDENCIES += $(SRCDIR)/helper-src/openssl/evp_override.goto
 
+REMOVE_FUNCTION_BODIES += --remove-function-body EVP_MD_CTX_free
+REMOVE_FUNCTION_BODIES += --remove-function-body evp_md_ctx_is_valid
```
danielsn, line 40: why

.cbmc-batch/stubs/EVP_MD_CTX_free_stub.c
```diff
@@ -0,0 +1,40 @@
...
+    bool is_initialized;
+    EVP_PKEY *pkey;
+    size_t digest_size;
+};
+
+/*
+ * Description: Cleans up digest context ctx and frees up the space allocated to it.
+ * Use this stub when we are *certain* there is no pkey associated with the digest context.
+ */
+void EVP_MD_CTX_free(EVP_MD_CTX *ctx) {
```
danielsn, line 35: Is there a ref counter to look at here?

.cbmc-batch/stubs/evp_md_ctx_is_valid_stub.c
```diff
@@ -0,0 +1,36 @@
...
+    bool is_initialized;
+    EVP_PKEY *pkey;
+    size_t digest_size;
+};
+
+/* Helper function for CBMC proofs: checks if EVP_MD_CTX is valid.
+ * Use this stub when we are *certain* there is no pkey associated with the digest context.
+ */
+bool evp_md_ctx_is_valid(EVP_MD_CTX *ctx) {
+    return ctx && ctx->is_initialized && ctx->digest_size <= EVP_MAX_MD_SIZE && ctx->pkey == NULL;
```
danielsn, line 35: why is null the only valid pkey

.cbmc-batch/jobs/aws_cryptosdk_md_abort/Makefile
tawdry-audrey: Does it cause the proof to hang to validate the evp_md_ctx due to lots of processing? If so this makes sense, just add a comment explaining that.

.cbmc-batch/stubs/EVP_MD_CTX_free_stub.c
tegansb: Not in this case, but if the pkey were not NULL there would be a ref counter to consider (as we do in evp_override.c).

.cbmc-batch/stubs/evp_md_ctx_is_valid_stub.c
tegansb: In general, it is not and the more generic evp_md_ctx_is_valid function reflects this.  Similarly as in https://github.com/aws/aws-encryption-sdk-c/pull/582#discussion_r478662276, I introduce this stub to improve coverage in the case where we are certain that pkey == NULL for the function being addressed in the CBMC proof. This greatly improves coverage. Once again, I have renamed these stubs to make this assumption more clear.

.cbmc-batch/jobs/aws_cryptosdk_md_abort/Makefile
tegansb: Same as in https://github.com/aws/aws-encryption-sdk-c/pull/582#discussion_r478660427. This has been updated!

feliperodri (changes requested):
I believe I posted all the comments for this PR in https://github.com/aws/aws-encryption-sdk-c/pull/581. Could you please address them here too?

.cbmc-batch/jobs/aws_cryptosdk_md_update/Makefile
```diff
@@ -17,7 +17,9 @@ sinclude ../Makefile.local
 #otherwise, use the default values
 include ../Makefile.local_default
 
-UNWINDSET += 
+# We abstract these functions as manual inspection shows there is no pkey associated with the MD_ctx.
```
danielsn, line 20: Is there an assert we can add to make sure that remains true?

danielsn (approved):

Issue #552, opened 2020-06-11
Tests should be executing the examples


robin-aws:
See #550 - one of the examples was compiling successfully but failing at runtime for quite a while without CI catching it.

PR #579, opened 2020-08-14
chore: run examples along with tests

alex-chew:
*Issue #, if available:* #552

*Description of changes:* This PR fixes some of the examples, and includes each example in the test suite.

Operational note: In order to run the tests that make KMS calls, it is necessary to set some environment variable to KMS key ARNs (e.g. `AWS_ENC_SDK_EXAMPLE_KEY_ARN`). This PR configured CMake to print an error to the user if any required environment variables are unset. I have set these environment variables appropriately for our CI environments, so that the CI runs for this PR should also run the examples.

# Check any applicable:


examples/CMakeLists.txt
```diff
@@ -30,6 +38,27 @@ if (BUILD_AWS_ENC_SDK_CPP)
...
+    aws_add_test(example_caching_cmm ${CMAKE_CURRENT_BINARY_DIR}/caching_cmm
+      $ENV{AWS_ENC_SDK_EXAMPLE_KEY_ARN} example hello world)
+    aws_add_test(example_file_streaming ${CMAKE_CURRENT_BINARY_DIR}/file_streaming
+      $ENV{AWS_ENC_SDK_EXAMPLE_KEY_ARN} ${CMAKE_CURRENT_SOURCE_DIR}/example_data/hello_world.txt)
+    aws_add_test(example_kms_discovery ${CMAKE_CURRENT_BINARY_DIR}/kms_discovery
+      $ENV{AWS_ENC_SDK_EXAMPLE_KEY_ARN_US_WEST_2} $ENV{AWS_ENC_SDK_EXAMPLE_KEY_ARN_EU_CENTRAL_1})
+    aws_add_test(example_multi_keyring ${CMAKE_CURRENT_BINARY_DIR}/multi_keyring
+      $ENV{AWS_ENC_SDK_EXAMPLE_KEY_ARN} ${CMAKE_CURRENT_SOURCE_DIR}/example_data/aes_256_key)
+    aws_add_test(example_string ${CMAKE_CURRENT_BINARY_DIR}/string
+      $ENV{AWS_ENC_SDK_EXAMPLE_KEY_ARN})
```
robin-aws, line 61: The tests that use KMS should only be added if AWS_ENC_SDK_END_TO_END_TESTS is enabled. That will be necessary to fix the Mac OS build at least.

The warning above about environment variables should probably also be inside that conditional.

examples/example_data/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.encrypted
```
robin-aws, line 1: Do you know which examples are spitting out these files, and can we fix them to clean up after themselves or have them write to a temporary directory? I hate messy tests. :)

robin-aws (dismissed):

examples/CMakeLists.txt
alex-chew: Ah, thanks for catching that.

examples/example_data/.gitignore
alex-chew: Yes, the `file_streaming.cpp` example generates `*.encrypted` and `*.decrypted` files alongside the original target file which is passed as an command-line argument. I'll look into how best to clean them up after tests are run.

PR #605, opened 2020-10-08
Update harnesses for aes_gcm_encrypt and aes_gcm_decrypt

tegansb:
*Description of changes:*
Added assumptions that the output buffer cannot be NULL. 
This assumption is needed. If the output buffer is NULL, the EVP_EncryptUpdate or EVP_DecryptUpdate stubs assume the case where the aad is being specified and do not update outl. This results in an overflow error as there are no constraints on the size of outl. The behavior of the stubs does seem to match the specification for the functions so I updated the assumptions instead of the stubs. 
Added matching preconditions to source code.

# Check any applicable:

### DIFF
--- a/source/cipher.c
+++ b/source/cipher.c
@@ -814,6 +814,13 @@ int aws_cryptosdk_aes_gcm_encrypt(
     const struct aws_byte_cursor iv,
     const struct aws_byte_cursor aad,
     const struct aws_string *key) {
+    AWS_PRECONDITION(aws_byte_buf_is_valid(cipher));
+    AWS_PRECONDITION(cipher->buffer != NULL);
+    AWS_PRECONDITION(aws_byte_buf_is_valid(tag));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&plain));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&iv));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&aad));
+    AWS_PRECONDITION(aws_string_is_valid(key));
     const EVP_CIPHER *alg = get_alg_from_key_size(key->len);
     if (!alg || iv.len != aes_gcm_iv_len || tag->capacity < aes_gcm_tag_len || cipher->capacity < plain.len)
         return aws_raise_error(AWS_ERROR_INVALID_BUFFER_SIZE);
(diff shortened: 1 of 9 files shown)


alex-chew (approved):
Overall LGTM. Do we have a mechanism to ensure that the preconditions in each operation under verification correspond to an assumption in the harness, or are we just doing that manually for now?

source/cipher.c
```diff
@@ -814,6 +814,13 @@ int aws_cryptosdk_aes_gcm_encrypt(
     const struct aws_byte_cursor iv,
     const struct aws_byte_cursor aad,
     const struct aws_string *key) {
+    AWS_PRECONDITION(aws_byte_buf_is_valid(cipher));
+    AWS_PRECONDITION(cipher->buffer != NULL);
+    AWS_PRECONDITION(aws_byte_buf_is_valid(tag));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&plain));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&iv));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&aad));
+    AWS_PRECONDITION(key != NULL);
```
feliperodri, line 823: Why not `AWS_PRECONDITION(aws_string_is_valid(key));`?

source/cipher.c
```diff
@@ -854,6 +861,13 @@ int aws_cryptosdk_aes_gcm_decrypt(
     const struct aws_byte_cursor iv,
     const struct aws_byte_cursor aad,
     const struct aws_string *key) {
+    AWS_PRECONDITION(aws_byte_buf_is_valid(plain));
+    AWS_PRECONDITION(plain->buffer != NULL);
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&cipher));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&tag));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&iv));
+    AWS_PRECONDITION(aws_byte_cursor_is_valid(&aad));
+    AWS_PRECONDITION(key != NULL);
```
feliperodri, line 870: Why not `AWS_PRECONDITION(aws_string_is_valid(key));`?

feliperodri (changes requested):
I have a couple questions. I also recommend to update the AWS C Common submodule to the latest version. @tegansb

source/cipher.c
tegansb: True that's a better check! Updated and also updated to latest c-common

source/cipher.c
tegansb: Same as above. Updated!

feliperodri (approved):
LGTM.

PR #602, opened 2020-10-08
Update aws_cryptodsk_sig_verify_start and aws_cryptodsk_sig_verify_finish harnesses

tegansb:
*Description of changes:*
Updates aws_cryptodsk_sig_verify_start and aws_cryptodsk_sig_verify_finish harnesses to handle new malloc-may-fail flag. 
Fixes harness assertion that should have been an assume. 
Matches preconditions in source code to that in harness. 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #606, opened 2020-10-09
Update CBMC harnesses for aws_cryptosdk_enc_ctx_serialize, aws_cryptosdk_sig_sign_start and aws_cryptosdk_sig_get_privkey

tegansb:
*Description of changes:*

Updates harnesses for aws_cryptosdk_enc_ctx_serialize, aws_cryptosdk_sig_sign_start and aws_cryptosdk_sig_get_privkey with explicit non-NULL assumptions about data structures. 
Added corresponding preconditions to functions in source code. 

# Check any applicable:


.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
```diff
@@ -27,24 +27,28 @@ void array_list_item_generator(struct aws_array_list *elems) {
     for (size_t index = 0; index < elems->length; ++index) {
         struct aws_hash_element *val = (struct aws_hash_element *)((uint8_t *)elems->data + (elems->item_size * index));
         // Due to the checks in aws_cryptosdk_enc_ctx_size, no string can have a length > UINT16_MAX
-        val->key   = ensure_string_is_allocated_bounded_length(UINT16_MAX);
+        __CPROVER_assume(val != NULL);
```
feliperodri, line 30: Don't we have an invariant for `aws_hash_element`?

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
```diff
@@ -27,24 +27,28 @@ void array_list_item_generator(struct aws_array_list *elems) {
...
+        val->key = ensure_string_is_allocated_bounded_length(UINT16_MAX);
+        __CPROVER_assume(val->key != NULL);
```
feliperodri, line 32: Why can't we assume here `aws_string_is_valid()`?

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
```diff
@@ -27,24 +27,28 @@ void array_list_item_generator(struct aws_array_list *elems) {
...
         val->value = ensure_string_is_allocated_bounded_length(UINT16_MAX);
+        __CPROVER_assume(val->value != NULL);
```
feliperodri, line 34: Why can't we assume here `aws_string_is_valid()`?

source/cipher_openssl.c
```diff
@@ -533,6 +534,7 @@ int aws_cryptosdk_sig_sign_start(
     AWS_PRECONDITION(AWS_OBJECT_PTR_IS_WRITABLE(ctx));
     AWS_PRECONDITION(AWS_OBJECT_PTR_IS_READABLE(alloc));
     AWS_PRECONDITION(AWS_OBJECT_PTR_IS_READABLE(props));
+    AWS_PRECONDITION(priv_key != NULL);
```
feliperodri, line 537: Why can't we assume here `aws_string_is_valid()`?

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
```diff
@@ -23,29 +23,30 @@
...
     struct aws_string *pub_key;
     struct aws_string *priv_key = ensure_string_is_allocated_nondet_length();
     enum aws_cryptosdk_alg_id alg_id;
     struct aws_cryptosdk_alg_properties *props = aws_cryptosdk_alg_props(alg_id);
 
-    /* assumptions */
-    __CPROVER_assume(props);
+    /* Assumptions */
+    __CPROVER_assume(props != NULL);
+    __CPROVER_assume(priv_key != NULL);
```
feliperodri, line 36: Why can't we assume here `aws_string_is_valid()`?

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
```diff
@@ -23,29 +23,30 @@
...
     struct aws_allocator *alloc = can_fail_allocator();
     struct aws_string *pub_key;
     struct aws_string *priv_key = ensure_string_is_allocated_nondet_length();
     enum aws_cryptosdk_alg_id alg_id;
     struct aws_cryptosdk_alg_properties *props = aws_cryptosdk_alg_props(alg_id);
 
-    /* assumptions */
-    __CPROVER_assume(props);
+    /* Assumptions */
+    __CPROVER_assume(props != NULL);
```
feliperodri, line 35: Why can't we assume here `aws_cryptosdk_alg_properties_is_valid()`?

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
```diff
@@ -23,28 +23,28 @@
...
-    /* arguments */
+    /* Nondet Input */
     struct aws_cryptosdk_sig_ctx *ctx = can_fail_malloc(sizeof(struct aws_cryptosdk_sig_ctx));
     struct aws_allocator *alloc       = can_fail_allocator();
     struct aws_string *priv_key;
 
-    /* assumptions */
-    __CPROVER_assume(ctx);
+    /* Assumptions */
+    __CPROVER_assume(ctx != NULL);
```
feliperodri, line 32: What we really want here is
```c
struct aws_cryptosdk_sig_ctx *ctx = ensure_sig_ctx_has_allocated_members();
```
And inside `ensure_sig_ctx_has_allocated_members` we nondeterministically allocate memory or `ctx`. I'd also expect that `aws_cryptosdk_sig_ctx_is_valid_cbmc` ensures `ctx` is different than NULL, which would avoid the need of the extra assumption. Why do we have a `cbmc` sufix in the name of that function?

source/cipher_openssl.c
```diff
@@ -323,6 +323,7 @@ int aws_cryptosdk_sig_get_pubkey(
 
 int aws_cryptosdk_sig_get_privkey(
     const struct aws_cryptosdk_sig_ctx *ctx, struct aws_allocator *alloc, struct aws_string **priv_key) {
+    AWS_PRECONDITION(ctx != NULL);
```
feliperodri, line 326: Why not assert `aws_cryptosdk_sig_ctx_is_valid_cbmc()`?

feliperodri (changes requested):

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
tegansb: I couldn't find one in c-common, either in hash_table.h or make_common_data_structures or other cbmc source code.

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
tegansb: Updated!

source/cipher_openssl.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
tegansb: Yep, updated as in https://github.com/aws/aws-encryption-sdk-c/pull/602.

source/cipher_openssl.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
```diff
@@ -27,24 +27,32 @@ void array_list_item_generator(struct aws_array_list *elems) {
...
+        __CPROVER_assume(value->len <= UINT16_MAX);
+        val->value = value;
     }
 }
 
 void aws_cryptosdk_enc_ctx_serialize_harness() {
-    // Precondition: byte_buf is valid
+    /* Nondet Input */
     struct aws_byte_buf output;
+    struct aws_hash_table map;
```
danielsn, line 45: make this a pointer and malloc it.

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
danielsn: Different hash elements will have different constraints. What you need is one that ensures the particular requirements of this type.

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
```diff
@@ -23,28 +23,26 @@
...
-    /* assumptions */
-    __CPROVER_assume(ctx);
-    ensure_sig_ctx_has_allocated_members(ctx);
+    /* Assumptions */
     __CPROVER_assume(aws_cryptosdk_sig_ctx_is_valid_cbmc(ctx));
     __CPROVER_assume(ctx->is_sign);  // context has to be in signing mode, otherwise private key is NULL
 
-    /* operation under verification */
+    /* Operation under verification */
     if (aws_cryptosdk_sig_get_privkey(ctx, alloc, &priv_key) == AWS_OP_SUCCESS) {
```
danielsn, line 36: Can `priv_key` be null?

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -61,7 +61,11 @@ void ensure_md_context_has_allocated_members(struct aws_cryptosdk_md_context *ct
     ctx->evp_md_ctx = evp_md_ctx_nondet_alloc();
 }
 
-void ensure_sig_ctx_has_allocated_members(struct aws_cryptosdk_sig_ctx *ctx) {
+struct aws_cryptosdk_sig_ctx *ensure_sig_ctx_has_allocated_members() {
+    struct aws_cryptosdk_sig_ctx *ctx = bounded_malloc(sizeof(*ctx));
```
danielsn, line 65: regular `malloc` works here

danielsn (changes requested):

.cbmc-batch/source/make_common_data_structures.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
tegansb: Yes, it can!

.cbmc-batch/source/make_common_data_structures.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
tegansb: Updated!

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
danielsn: But `&priv_key` will always point to something. Is that what we want?

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
```diff
@@ -23,29 +23,29 @@
...
     struct aws_string *priv_key = ensure_string_is_allocated_nondet_length();
     enum aws_cryptosdk_alg_id alg_id;
     struct aws_cryptosdk_alg_properties *props = aws_cryptosdk_alg_props(alg_id);
 
-    /* assumptions */
-    __CPROVER_assume(props);
-    assert(aws_string_is_valid(priv_key));
+    /* Assumptions */
+    __CPROVER_assume(aws_cryptosdk_alg_properties_is_valid(NULL));
+    __CPROVER_assume(aws_string_is_valid(NULL));
```
danielsn, line 36: Is this what you meant????

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
```diff
@@ -23,29 +23,29 @@
...
 
     bool save_pub_key = nondet_bool();
 
-    /* operation under verification */
+    /* Operation under verification */
     if (aws_cryptosdk_sig_sign_start(
             &ctx, alloc, save_pub_key ? &pub_key : NULL, props, priv_key /* priv_key can't be NULL */) ==
```
danielsn, line 42: `&ctx` is never null. Is this what you intend?

danielsn (changes requested):

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
tegansb: LOL, definitely not. My brain must just be fixated on the word "NULL". Fixed now.

feliperodri (changes requested):

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
danielsn: We're taking `priv_key` as a `**`, so we need to check if either level of the indirection is null

.cbmc-batch/jobs/aws_cryptosdk_sig_sign_start/aws_cryptosdk_sig_sign_start_harness.c
tegansb: ctx itself can be NULL, I've updated the harness. Its address shouldn't be.

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
tegansb: I've allocated priv_key like so  
 `  struct aws_string *priv_key       = ensure_string_is_allocated_nondet_length();`
and then assumed that it's either NULL or a valid string and added that precondition to the source code. priv_key is immediately set to NULL in the function 
`*priv_key = NULL;`. 
However, when I allocate this way or through malloc() as above, I get a memory leak error even though I call free(priv_key) at the end of the function. I'm not very used to using the  --memory-leak-check flag, so any idea why this might be?

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
danielsn: I'm concerned about this `val != NULL` assumption. val comes from `data` so if its `null`, something has gone wrong higher up in the code. I'd prefer to make this `assert(AWS_MEM_IS_WRITABLE(val, elems->item_size)`

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
```diff
@@ -23,28 +23,27 @@
 #include <cipher_openssl.h>
 
 void aws_cryptosdk_sig_get_privkey_harness() {
-    /* arguments */
-    struct aws_cryptosdk_sig_ctx *ctx = can_fail_malloc(sizeof(struct aws_cryptosdk_sig_ctx));
+    /* Nondet Input */
+    struct aws_cryptosdk_sig_ctx *ctx = ensure_nondet_sig_ctx_has_allocated_members();
     struct aws_allocator *alloc       = can_fail_allocator();
-    struct aws_string *priv_key;
+    struct aws_string *priv_key       = ensure_string_is_allocated_nondet_length();
```
danielsn, line 29: This is OK for this PR, but we should really rename the `ensure` functions to make it clear that they may or may not allocate the fields.

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
danielsn: I think the memory-leak check is not worth it, personally.

danielsn (changes requested):

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
tegansb: So I just tried out that assert and it fails. I think it's because we don't actually specify anything about the elements of the hash table, and we use this array_list_item_generator as a generator for those elements. I found this comment describing the strategy: 
```
 /* If the consumer of the list doesn't use the elements in the list, we can just leave it undef
 * This is sound, as it gives you a totally nondet value every time you use a list element, and is the default behaviour
 * of CBMC. But if it is used, we need a way for the harness to specify valid values for the element, for example if
 * they are copying values out of the table. They can do this by defining
 * -DAWS_CRYPTOSDK_HASH_ELEMS_ARRAY_INIT_GENERATOR=the_generator_fn, wher the_generator_fn has signature
 * the_generator_fn(struct aws_array_list *elems).
 *   [elems] is a pointer to the array_list whose values need to be set
 */
```

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
tegansb: I've made github issues for those both in ESDK (https://github.com/aws/aws-encryption-sdk-c/issues/614) and c-common (https://github.com/awslabs/aws-c-common/issues/723)

.cbmc-batch/jobs/aws_cryptosdk_sig_get_privkey/aws_cryptosdk_sig_get_privkey_harness.c
tegansb: I've removed it.

.cbmc-batch/jobs/aws_cryptosdk_enc_ctx_serialize/aws_cryptosdk_enc_ctx_serialize_harness.c
danielsn: If that assertion fails, that implies that whatever setup is being done for the array list is insufficient.

tegansb: I'm looking into this. The array list isn't set up as part of the harness but rather through aws_cryptosdk_hash_elems_array_init which we stub out (to handle CBMC function pointer removal). The stub for aws_cryptosdk_hash_elems_array_init uses the generator defined in the harness to nondet specify elements (with the assumptions from the harness on them). This is why the assert fails. 
I'm trying our replacing the stub with the actual function but elements are copied in via memset and other functions needed for the assertions to hold are also stubbed out, so it's a bit of a mess. Hopefully I can get something sorted out.

feliperodri (approved):
Despite the dsn's comment on the assumption about `val`, LGTM.

danielsn (approved):

alex-chew (approved):

PR #607, opened 2020-10-09
Updates to aws_atomic_var_is_valid and aws_cryptosdk_keyring_base_init harness

tegansb:
*Description of changes:*
Updates the harness of aws_cryptosdk_keyring_base_init with non-NULL assumptions over vtable->name.
Modifies the aws_atomic_var_is_valid check to be less strict. Unsure if this change is desired yet, still looking into it. 

# Check any applicable:

### DIFF
--- a/.cbmc-batch/include/make_common_data_structures.h
+++ b/.cbmc-batch/include/make_common_data_structures.h
@@ -48,4 +48,4 @@ void ensure_trace_has_allocated_records(struct aws_array_list *trace, size_t max
 void ensure_cryptosdk_keyring_has_allocated_members(
     struct aws_cryptosdk_keyring *keyring, const struct aws_cryptosdk_keyring_vt *vtable);
 /* Non-deterministically allocates a aws_cryptosdk_keyring_vt structure with a valid name*/
-void ensure_vtable_has_allocated_members(struct aws_cryptosdk_keyring_vt *vtable, size_t max_len);
+void ensure_nondet_allocate_vtable_members(struct aws_cryptosdk_keyring_vt *vtable, size_t max_len);
--- a/.cbmc-batch/include/make_common_data_structures.h
+++ b/.cbmc-batch/include/make_common_data_structures.h
@@ -47,3 +47,5 @@ void ensure_trace_has_allocated_records(struct aws_array_list *trace, size_t max
 /* Non-deterministically allocates a aws_cryptosdk_keyring structure */
 void ensure_cryptosdk_keyring_has_allocated_members(
     struct aws_cryptosdk_keyring *keyring, const struct aws_cryptosdk_keyring_vt *vtable);
+/* Non-deterministically allocates a aws_cryptosdk_keyring_vt structure with a valid name*/
+void ensure_vtable_has_allocated_members(struct aws_cryptosdk_keyring_vt *vtable, size_t max_len);
--- a/.cbmc-batch/include/make_common_data_structures.h
+++ b/.cbmc-batch/include/make_common_data_structures.h
@@ -48,4 +48,6 @@ void ensure_trace_has_allocated_records(struct aws_array_list *trace, size_t max
 void ensure_cryptosdk_keyring_has_allocated_members(
     struct aws_cryptosdk_keyring *keyring, const struct aws_cryptosdk_keyring_vt *vtable);
 /* Non-deterministically allocates a aws_cryptosdk_keyring_vt structure with a valid name*/
-void ensure_nondet_allocate_vtable_members(struct aws_cryptosdk_keyring_vt *vtable, size_t max_len);
+void ensure_nondet_allocate_keyring_vtable_members(struct aws_cryptosdk_keyring_vt *vtable, size_t max_len);
+/* Non-deterministically allocates a aws_cryptosdk_cmm_vt structure with a valid name*/
+void ensure_nondet_allocate_cmm_vtable_members(struct aws_cryptosdk_cmm_vt *vtable, size_t max_len);
--- a/.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
+++ b/.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
@@ -23,10 +23,17 @@
 #include <proof_helpers/utils.h>
 
 void aws_cryptosdk_cmm_base_init_harness() {
+    /* Nondet input */
     struct aws_cryptosdk_cmm cmm;
     const struct aws_cryptosdk_cmm_vt vtable;
+
+    /* Assumptions */
     *(char **)(&vtable.name) = ensure_c_str_is_allocated(SIZE_MAX);
     __CPROVER_assume(aws_cryptosdk_cmm_vtable_is_valid(&vtable));
+
+    /* Operation under verification */
     aws_cryptosdk_cmm_base_init(&cmm, &vtable);
+
+    /* Post-conditions */
     assert(aws_cryptosdk_cmm_base_is_valid(&cmm));
 }
--- a/.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
+++ b/.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
@@ -28,7 +28,7 @@ void aws_cryptosdk_cmm_base_init_harness() {
     const struct aws_cryptosdk_cmm_vt vtable;
 
     /* Assumptions */
-    ensure_vtable_has_allocated_members(&vtable, SIZE_MAX);
+    ensure_nondet_allocate_vtable_members(&vtable, SIZE_MAX);
     __CPROVER_assume(aws_cryptosdk_cmm_vtable_is_valid(&vtable));
 
     /* Operation under verification */
--- a/.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
+++ b/.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
@@ -17,7 +17,7 @@
  */
 
 #include <aws/cryptosdk/materials.h>
-#include <proof_helpers/cryptosdk/make_common_data_structures.h>
+#include <make_common_data_structures.h>
 #include <proof_helpers/make_common_data_structures.h>
 #include <proof_helpers/proof_allocators.h>
 #include <proof_helpers/utils.h>
@@ -28,7 +28,7 @@ void aws_cryptosdk_cmm_base_init_harness() {
     const struct aws_cryptosdk_cmm_vt vtable;
 
     /* Assumptions */
-    *(char **)(&vtable.name) = ensure_c_str_is_allocated(SIZE_MAX);
+    ensure_vtable_has_allocated_members(&vtable, SIZE_MAX);
     __CPROVER_assume(aws_cryptosdk_cmm_vtable_is_valid(&vtable));
 
     /* Operation under verification */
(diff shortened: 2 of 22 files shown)


.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
```diff
@@ -17,12 +17,15 @@
...
 
 void aws_cryptosdk_keyring_base_init_harness() {
-    /* Non-deterministic inputs. */
+    /* Nondet input. */
     struct aws_cryptosdk_keyring *keyring   = can_fail_malloc(sizeof(*keyring));
     struct aws_cryptosdk_keyring_vt *vtable = can_fail_malloc(sizeof(*vtable));
 
-    /* Pre-conditions. */
+    /* Assumptions. */
     __CPROVER_assume(keyring != NULL);
```
feliperodri, line 25: Don't we have a `ensure_cryptosdk_keyring_is_allocated` function? What about an invariant (i.e. `_is_valid` function)? I'd prefer using them here instead of malloc, followed by this assumption. Also, this function name would not even be appropriate anymore, because these predicates should not ensure allocation, but instead nondeterministically allocate memory, so they should be renamed to something like `nondet_allocate_cryptosdk_keyring`.

.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
```diff
@@ -17,12 +17,15 @@
...
+    __CPROVER_assume(IMPLIES(vtable != NULL, vtable->name = ensure_c_str_is_allocated(MAX_STRING_LEN)));
```
feliperodri, line 26: Same comment as above.

.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
```diff
@@ -17,12 +17,15 @@
...
+    __CPROVER_assume(IMPLIES(vtable != NULL, vtable->name != NULL));
+    __CPROVER_assume(IMPLIES(vtable != NULL, aws_c_string_is_valid(vtable->name)));
```
feliperodri, line 28: If you have this assumption here, why the previous one? I assume a valid string woundn't be null.

include/aws/cryptosdk/materials.h
```diff
@@ -400,7 +400,14 @@ struct aws_cryptosdk_cmm_vt {
...
  * Putting this here for now, until we get it merged into the atomics.h in c-common
  */
-AWS_CRYPTOSDK_STATIC_INLINE bool aws_atomic_var_is_valid(const struct aws_atomic_var *var) {
+AWS_CRYPTOSDK_STATIC_INLINE bool aws_atomic_var_is_valid_int(const struct aws_atomic_var *var) {
+    return AWS_MEM_IS_WRITABLE(var, sizeof(size_t));
+}
+
+/**
+ * Putting this here for now, until we get it merged into the atomics.h in c-common
+ */
```
feliperodri, line 409: Do we have a GitHub issue to track this? It would be nice to get @nchong-at-aws input here too.

.cbmc-batch/jobs/aws_cryptosdk_keyring_on_decrypt/aws_cryptosdk_keyring_on_decrypt_harness.c
```diff
@@ -82,6 +82,7 @@ void aws_cryptosdk_keyring_on_decrypt_harness() {
                                                      .on_decrypt = nondet_bool() ? NULL : on_decrypt };
     struct aws_cryptosdk_keyring keyring;
     ensure_cryptosdk_keyring_has_allocated_members(&keyring, &vtable);
+    __CPROVER_assume(keyring.refcount.value != NULL);
```
feliperodri, line 85: Why not assume the invariant here? What is the equivalence precondition in the harness?

.cbmc-batch/jobs/aws_cryptosdk_keyring_on_encrypt/aws_cryptosdk_keyring_on_encrypt_harness.c
```diff
@@ -89,6 +89,7 @@ void aws_cryptosdk_keyring_on_encrypt_harness() {
                                                      .on_decrypt = nondet_voidp() };
     struct aws_cryptosdk_keyring keyring;
     ensure_cryptosdk_keyring_has_allocated_members(&keyring, &vtable);
+    __CPROVER_assume(keyring.refcount.value != NULL);
```
feliperodri, line 92: Why not assume the invariant here? What is the equivalence precondition in the harness?

feliperodri (changes requested):

.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
tegansb: Fixed! Created a ensure_vtable_has_allocated_members function for this.

.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
tegansb: Yep, fixed! The non-NULL assumption is not needed.

include/aws/cryptosdk/materials.h
tegansb: Yep, we do! https://github.com/aws/aws-encryption-sdk-c/issues/612

.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
tegansb: We do have a ensure_cryptosdk_keyring_is_allocated but the function under test (aws_cryptosdk_keyring_base_init) is actually the function that initializes a keyring (including giving it its vtable which is also what the ensure_cryptosdk_keyring_is_allocated does). I changed the can_fail_malloc to bounded_malloc though.

.cbmc-batch/jobs/aws_cryptosdk_keyring_on_decrypt/aws_cryptosdk_keyring_on_decrypt_harness.c
tegansb: This assumption is actually no longer needed. It was a weird property of a stub we were using but we can actually use the real function so I've updated to do that. The aws_cryptosdk_keyring_is_valid invariant is enough and the condition in the harness matches.

.cbmc-batch/jobs/aws_cryptosdk_keyring_on_encrypt/aws_cryptosdk_keyring_on_encrypt_harness.c
tegansb: Same as above!

.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
```diff
@@ -23,10 +23,17 @@
 #include <proof_helpers/utils.h>
 
 void aws_cryptosdk_cmm_base_init_harness() {
+    /* Nondet input */
     struct aws_cryptosdk_cmm cmm;
     const struct aws_cryptosdk_cmm_vt vtable;
+
+    /* Assumptions */
     *(char **)(&vtable.name) = ensure_c_str_is_allocated(SIZE_MAX);
```
danielsn, line 31: why is this not part of `ensure_vtable_has_allocated_members`

.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
```diff
@@ -17,13 +17,14 @@
 #include <make_common_data_structures.h>
 
 void aws_cryptosdk_keyring_base_init_harness() {
-    /* Non-deterministic inputs. */
-    struct aws_cryptosdk_keyring *keyring   = can_fail_malloc(sizeof(*keyring));
-    struct aws_cryptosdk_keyring_vt *vtable = can_fail_malloc(sizeof(*vtable));
+    /* Nondet input. */
+    struct aws_cryptosdk_keyring *keyring   = bounded_malloc(sizeof(*keyring));
+    struct aws_cryptosdk_keyring_vt *vtable = bounded_malloc(sizeof(*vtable));
```
danielsn, line 22: Why not just `malloc`?

danielsn (changes requested):

.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c
tegansb: It is, updated to use this function!

.cbmc-batch/jobs/aws_cryptosdk_keyring_base_init/aws_cryptosdk_keyring_base_init_harness.c
tegansb: Updated!

.cbmc-batch/source/make_common_data_structures.c
```diff
@@ -152,3 +152,10 @@ void ensure_cryptosdk_keyring_has_allocated_members(
     keyring->refcount.value = can_fail_malloc(sizeof(size_t));
     keyring->vtable         = nondet_bool() ? NULL : vtable;
 }
+
+void ensure_vtable_has_allocated_members(struct aws_cryptosdk_keyring_vt *vtable, size_t max_len) {
+    if (vtable) {
+        vtable->name = ensure_c_str_is_allocated(max_len);
+        __CPROVER_assume(aws_c_string_is_valid(vtable->name));
+    }
+}
```
feliperodri, line 161: I'd rename this function to `nondet_allocate_vtable_members` and move the assumption to a validity function (e.g., `aws_cryptosdk_keyring_vt_is_valid`). Here an example [ensure_c_str_is_allocated](https://github.com/awslabs/aws-c-common/blob/e4d885544ae1f80a2e802d7779e3e754e5562fd0/verification/cbmc/sources/make_common_data_structures.c#L191).

feliperodri (changes requested):

.cbmc-batch/source/make_common_data_structures.c
tegansb: Updated!

feliperodri (approved):
LGTM.

danielsn (approved):

.cbmc-batch/include/make_common_data_structures.h
```diff
@@ -47,3 +47,5 @@ void ensure_trace_has_allocated_records(struct aws_array_list *trace, size_t max
 /* Non-deterministically allocates a aws_cryptosdk_keyring structure */
 void ensure_cryptosdk_keyring_has_allocated_members(
     struct aws_cryptosdk_keyring *keyring, const struct aws_cryptosdk_keyring_vt *vtable);
+/* Non-deterministically allocates a aws_cryptosdk_keyring_vt structure with a valid name*/
+void ensure_nondet_allocate_vtable_members(struct aws_cryptosdk_keyring_vt *vtable, size_t max_len);
```
alex-chew, line 51: This name seems like it would lead to confusion between keyring vtables and CMM vtables. For example in `.cbmc-batch/jobs/aws_cryptosdk_cmm_base_init/aws_cryptosdk_cmm_base_init_harness.c` line 31, you've passed a CMM vtable pointer where the function declaration suggests it should be a keyring vtable pointer. I assume this works out since `ensure_nondet_allocate_vtable_members` only accesses the `name` fields of the vtables, which happen to be identically positioned in the respective structs, but I'd prefer to avoid relying on that.

alex-chew (approved):
LGTM

.cbmc-batch/include/make_common_data_structures.h
tegansb: Good point! I've updated and made two separate functions, one for cmm and one for keyring_trace. They do the same thing right now but who knows, maybe down the line they'll diverge. Just waiting for the checks to go through now.

Issue #634, opened 2020-11-20
Use safe memset in derive_key_v2


alex-chew:
In `aws_cryptosdk_private_derive_key_v2` we write:

https://github.com/aws/aws-encryption-sdk-c/blob/939c422c10c5984b4e86a551c71910d384d4debe/source/cipher.c#L360

with the intention that `props->commitment_len` may be zero, in which case `commitment->buffer` is not writable, but this is OK since the `memset` would be a no-op. The C standard says that even in this case the pointer argument `commitment->buffer` must be valid. We should change this so that `memset` is not called with an invalid pointer argument.

PR #635, opened 2020-11-21
Only memset with valid pointer in derive_key_v2

alex-chew:
*Issue #, if available:* #634 

*Description of changes:* Only memset with valid pointer in derive_key_v2

# Check any applicable:

### DIFF
--- a/source/cipher.c
+++ b/source/cipher.c
@@ -358,7 +358,7 @@ static int aws_cryptosdk_private_derive_key_v2(
 
     memset(content_key->keybuf, 0, sizeof(content_key->keybuf));
     if (props->commitment_len) {
-        assert(AWS_MEM_IS_WRITABLE(commitment->buffer, props->commitment_len));
+        assert(commitment->buffer);
         memset(commitment->buffer, 0, props->commitment_len);
     }
 
--- a/source/cipher.c
+++ b/source/cipher.c
@@ -357,7 +357,10 @@ static int aws_cryptosdk_private_derive_key_v2(
     const struct aws_byte_buf myikm  = aws_byte_buf_from_array(data_key->keybuf, props->data_key_len);
 
     memset(content_key->keybuf, 0, sizeof(content_key->keybuf));
-    memset(commitment->buffer, 0, props->commitment_len);
+    if (props->commitment_len) {
+        assert(AWS_MEM_IS_WRITABLE(commitment->buffer, props->commitment_len));
+        memset(commitment->buffer, 0, props->commitment_len);
+    }
 
     uint8_t derivekey_info_array[] = "\0\0DERIVEKEY";
     derivekey_info_array[0]        = props->alg_id >> 8;


source/cipher.c
```diff
@@ -357,7 +357,10 @@ static int aws_cryptosdk_private_derive_key_v2(
     const struct aws_byte_buf myikm  = aws_byte_buf_from_array(data_key->keybuf, props->data_key_len);
 
     memset(content_key->keybuf, 0, sizeof(content_key->keybuf));
-    memset(commitment->buffer, 0, props->commitment_len);
+    if (props->commitment_len) {
+        assert(AWS_MEM_IS_WRITABLE(commitment->buffer, props->commitment_len));
```
feliperodri, line 361: In this case,
```c
#    define AWS_MEM_IS_WRITABLE(base, len) (((len) == 0) || (base))
```
So if `props->commitment_len` is 0, we can still pass a NULL pointer to `memset`. I'd recommend the following:
```c
assert(commitment->buffer && AWS_MEM_IS_WRITABLE(commitment->buffer, props->commitment_len));
```

feliperodri (changes requested):

source/cipher.c
adpaco-aws: But if `props->commitment_len` is 0, neither `assert` or `memset` are executed. Right?

In this case, `AWS_MEM_IS_WRITABLE(base, len)` is equivalent to `base`, so we can just write:
```
assert(commitment->buffer);
``` 

Thanks for bringing this up!

alex-chew: Thanks all. I agree with Adrian here - the guard ensures that the assert is only reached when `props->commitment_len` is positive.

However, if I understand correctly, `AWS_MEM_IS_WRITABLE(base, len)` might not be equivalent to `base` when running under CBMC:

```c
/**
 * Define function contracts.
 * When the code is being verified using CBMC these contracts are formally verified;
 * When the code is built in debug mode, they are checked as much as possible using assertions
 * When the code is built in production mode, non-fatal contracts are not checked.
 * Violations of the function contracts are undefined behaviour.
 */
#ifdef CBMC
// snip
#    define AWS_MEM_IS_READABLE(base, len) (((len) == 0) || (__CPROVER_r_ok((base), (len))))
#    define AWS_MEM_IS_WRITABLE(base, len) (((len) == 0) || (__CPROVER_r_ok((base), (len))))
#else
// snip
/**
 * The C runtime does not give a way to check these properties,
 * but we can at least check that the pointer is valid. */
#    define AWS_MEM_IS_READABLE(base, len) (((len) == 0) || (base))
#    define AWS_MEM_IS_WRITABLE(base, len) (((len) == 0) || (base))
#endif /* CBMC */
```

(from https://github.com/awslabs/aws-c-common/blob/a61f2ea5fbc9ac2eee9fe32f7a3fdbd399d9fbdb/include/aws/common/assert.h#L73)

Can I safely simplify the `AWS_MEM_IS_WRITABLE(base, len)` call to just `base` in the both cases?

feliperodri: You're both right. I believe in this case, the best option would be
```c
assert(commitment->buffer);
```

alex-chew: Got it, I'll change that then.

feliperodri (approved):

PR #632, opened 2020-11-18
Adds a CBMC proof for aws_cryptosdk_hdr_size function

tegansb:
*Description of changes:*
Proof harness and Makefile for aws_cryptosdk_hdr_size. 
Adds new validity, is_bounded and ensures function for the aws_cryptosdk_hdr struct. 
Adds preconditions to the aws_cryptosdk_hdr_size function. 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #649, opened 2020-12-03
Adds CBMC proof for aws_cryptosdk_priv_hdr_parse_edks

tegansb:
Adds harness, Makefile and yaml for aws_cryptosdk_priv_hdr_parse_edks.

Uses ensures, is bounded and validity functions from #632.

# Check any applicable:


alex-chew (approved):
LGTM

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/aws_cryptosdk_priv_hdr_parse_edks_harness.c
```diff
@@ -0,0 +1,81 @@
...
+    return rval;
+}
+
+void aws_cryptosdk_priv_hdr_parse_edks_harness() {
+    /* Nondet Input */
+    struct aws_cryptosdk_hdr *hdr   = hdr_setup(MAX_TABLE_SIZE, MAX_EDK_LIST_ITEMS, MAX_BUFFER_SIZE);
+    struct aws_byte_cursor *pcursor = malloc(sizeof(*pcursor));
+
+    /* Assumptions */
+    __CPROVER_assume(&hdr->alloc != NULL);
```
feliperodri, line 45: You don't need `&hdr->alloc`, right? Why not just `hdr->alloc != NULL`?

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/aws_cryptosdk_priv_hdr_parse_edks_harness.c
```diff
@@ -0,0 +1,81 @@
...
+uint16_t aws_byte_cursor_read_be16_generator_for_parse_edks(const struct aws_byte_cursor *cursor) {
+    (void)cursor;
+    static int num_times_called = 0;
+    num_times_called++;
+    uint16_t rval;
+    if (num_times_called == 1) {
+        __CPROVER_assume(rval <= MAX_EDK_LIST_ITEMS);
+    }
+    return rval;
+}
```
feliperodri, line 37: nit: I'd suggest:

```c
static bool flag = true;

/**
 * In the aws_cryptosdk_enc_ctx_deserilize() proof, the first value we read is the number of elements,
 * which we need to be constrained in order to ensure that the proof finishes. All other values can be left nondet.
 * This generates exactly that set of bytes
 */
uint16_t aws_byte_cursor_read_be16_generator_for_parse_edks(const struct aws_byte_cursor *cursor) {
    (void)cursor;
    uint16_t rval;
    if (flag) {
        __CPROVER_assume(rval <= MAX_EDK_LIST_ITEMS);
        flag = false; 
    }
    return rval;
}
```

feliperodri (changes requested):

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/Makefile
```diff
@@ -0,0 +1,83 @@
...
+DEFINES += -DAWS_BYTE_CURSOR_READ_BE16_GENERATOR=aws_byte_cursor_read_be16_generator_for_parse_edks
+DEFINES += -DAWS_NO_STATIC_IMPL
+DEFINES += -DMAX_EDK_LIST_ITEMS=$(MAX_EDK_LIST_ITEMS)
+DEFINES += -DMAX_TABLE_SIZE=$(MAX_TABLE_SIZE)
+
+# The maximum number of objects is set to 2^8 = 256. This option changes that to 2^9. 
+# Without this flag, running this proof causes the CBMC error of 
+# "too many addressed objects: maximum number of objects is set to 2^n=256 (with n=8);
+# use the `--object-bits n` option to increase the maximum number"
+CBMCFLAGS += --object-bits 9
```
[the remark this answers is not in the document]
adpaco-aws: There is a CBMC_OBJECT_BITS variable in Makefile.common that I think you should use here instead. I am afraid you may be running into `cbmc --object-bits 8 --object-bits 9` at the moment which could cause problems in the future.

adpaco-aws (changes requested):

source/header.c
```diff
@@ -286,6 +286,9 @@ int aws_cryptosdk_priv_hdr_parse_aad(struct aws_cryptosdk_hdr *hdr, struct aws_b
 }
 
 int aws_cryptosdk_priv_hdr_parse_edks(struct aws_cryptosdk_hdr *hdr, struct aws_byte_cursor *cur) {
+    AWS_PRECONDITION(aws_cryptosdk_hdr_is_valid(hdr));
+    AWS_PRECONDITION(&hdr->alloc != NULL);
```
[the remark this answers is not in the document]
tegansb: I actually don't need either, as I've since added the validity check on the allocator. I've removed this assumption.

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/aws_cryptosdk_priv_hdr_parse_edks_harness.c
tegansb: See above.

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/aws_cryptosdk_priv_hdr_parse_edks_harness.c
tegansb: Updated!

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/Makefile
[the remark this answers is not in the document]
tegansb: I think because of other changes in this proof I no longer need to change the object-bits option. I think I still will for sign_header and priv_try_gen_key, so I will look into what is causing the need for it if that is so.

adpaco-aws (approved):
The proof takes 2 hours to complete but seems correct. We should consider if there are any places where we can speed up the proof.

tegansb:
> The proof takes 2 hours to complete but seems correct. We should consider if there are any places where we can speed up the proof.

Sounds good! I'll think about what I can try and stub out. It's comforting to know the proof works as is with minimal stubbing but it's a long proof to wait for.

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/Makefile
```diff
@@ -0,0 +1,77 @@
...
+HARNESS_ENTRY = $(PROOF_UID)_harness
+HARNESS_FILE = $(HARNESS_ENTRY).c
+
+# Values are chosen for performance. Increasing values does not improve coverage.
+MAX_EDK_LIST_ITEMS ?= 1
+MAX_TABLE_SIZE ?= 2
+
+DEFINES += -DARRAY_LIST_TYPE="struct aws_cryptosdk_edk"
+DEFINES += -DARRAY_LIST_TYPE_HEADER=\"aws/cryptosdk/edk.h\"
+DEFINES += -DAWS_BYTE_CURSOR_READ_BE16_GENERATOR=aws_byte_cursor_read_be16_generator_for_parse_edks
```
feliperodri, line 31: Could you open a GitHub issue to improve the documentation of these DEFINES across all proofs?

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/aws_cryptosdk_priv_hdr_parse_edks_harness.c
```diff
@@ -0,0 +1,80 @@
...
+#include <make_common_data_structures.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/proof_allocators.h>
+#include <proof_helpers/utils.h>
+
+static bool flag = true;
+
+/**
+ * In the aws_cryptosdk_enc_ctx_deserilize() proof, the first value we read is the number of elements,
+ * which we need to be constrained in order to ensure that the proof finishes. All other values can be left nondet.
```
feliperodri, line 27: `aws_cryptosdk_enc_ctx_deserialize` and `which needs to be constrained in order...`

feliperodri (approved):
LGTM modulo a few typos.

alex-chew (approved):
LGTM

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/Makefile
tegansb: Done!

verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_edks/aws_cryptosdk_priv_hdr_parse_edks_harness.c
tegansb: Updated!

PR #656, opened 2020-12-04
Add CBMC proof for default_cmm_generate_enc_materials

adpaco-aws:
This PR adds a CBMC proof for default_cmm_generate_enc_materials, a function from default_cmm.c,

# Check any applicable:

### DIFF
--- a/source/default_cmm.c
+++ b/source/default_cmm.c
@@ -187,7 +187,8 @@ int aws_cryptosdk_default_cmm_set_alg_id(struct aws_cryptosdk_cmm *cmm, enum aws
 }
 
 bool aws_cryptosdk_default_cmm_is_valid(const struct aws_cryptosdk_cmm *cmm) {
-    AWS_PRECONDITION(cmm != NULL);
+    if (cmm == NULL) return false;
+
     struct default_cmm *self = (struct default_cmm *)cmm;
     return aws_cryptosdk_cmm_base_is_valid(&self->base) && aws_allocator_is_valid(self->alloc) &&
            aws_cryptosdk_keyring_is_valid(self->kr);
--- a/source/default_cmm.c
+++ b/source/default_cmm.c
@@ -38,7 +38,8 @@ static int default_cmm_generate_enc_materials(
     struct aws_cryptosdk_enc_materials **output,
     struct aws_cryptosdk_enc_request *request) {
     AWS_PRECONDITION(aws_cryptosdk_default_cmm_is_valid(cmm));
-    AWS_PRECONDITION(aws_cryptosdk_enc_materials_is_valid(output));
+    AWS_PRECONDITION(output != NULL);
+    AWS_PRECONDITION(aws_cryptosdk_enc_materials_is_valid(*output));
     AWS_PRECONDITION(aws_cryptosdk_enc_request_is_valid(request));
 
     struct aws_cryptosdk_enc_materials *enc_mat = NULL;
--- a/source/default_cmm.c
+++ b/source/default_cmm.c
@@ -39,7 +39,6 @@ static int default_cmm_generate_enc_materials(
     struct aws_cryptosdk_enc_request *request) {
     AWS_PRECONDITION(aws_cryptosdk_default_cmm_is_valid(cmm));
     AWS_PRECONDITION(output != NULL);
-    AWS_PRECONDITION(aws_cryptosdk_enc_materials_is_valid(*output));
     AWS_PRECONDITION(aws_cryptosdk_enc_request_is_valid(request));
 
     struct aws_cryptosdk_enc_materials *enc_mat = NULL;
--- a/source/default_cmm.c
+++ b/source/default_cmm.c
@@ -29,7 +29,7 @@ struct default_cmm {
     struct aws_cryptosdk_cmm base;
     struct aws_allocator *alloc;
     struct aws_cryptosdk_keyring *kr;
-    // Invariant: this is either DEFAULT_ALG_UNSET or is a valid algorithm ID
+    /* Invariant: this is either DEFAULT_ALG_UNSET or is a valid algorithm ID */
     enum aws_cryptosdk_alg_id default_alg;
 };
 
--- a/source/default_cmm.c
+++ b/source/default_cmm.c
@@ -37,8 +37,8 @@ static int default_cmm_generate_enc_materials(
     struct aws_cryptosdk_cmm *cmm,
     struct aws_cryptosdk_enc_materials **output,
     struct aws_cryptosdk_enc_request *request) {
-    AWS_PRECONDITION(cmm != NULL);
-    AWS_PRECONDITION(output != NULL);
+    AWS_PRECONDITION(aws_cryptosdk_default_cmm_is_valid(cmm));
+    AWS_PRECONDITION(aws_cryptosdk_enc_materials_is_valid(output));
     AWS_PRECONDITION(aws_cryptosdk_enc_request_is_valid(request));
 
     struct aws_cryptosdk_enc_materials *enc_mat = NULL;
--- a/source/default_cmm.c
+++ b/source/default_cmm.c
@@ -37,6 +37,10 @@ static int default_cmm_generate_enc_materials(
     struct aws_cryptosdk_cmm *cmm,
     struct aws_cryptosdk_enc_materials **output,
     struct aws_cryptosdk_enc_request *request) {
+    AWS_PRECONDITION(cmm != NULL);
+    AWS_PRECONDITION(output != NULL);
+    AWS_PRECONDITION(aws_cryptosdk_enc_request_is_valid(request));
+
     struct aws_cryptosdk_enc_materials *enc_mat = NULL;
     struct default_cmm *self                    = (struct default_cmm *)cmm;
     struct aws_hash_element *pElement           = NULL;
@@ -59,6 +63,7 @@ static int default_cmm_generate_enc_materials(
         }
     }
     const struct aws_cryptosdk_alg_properties *props = aws_cryptosdk_alg_props(request->requested_alg);
+    if (!props) goto err;
 
     enc_mat = aws_cryptosdk_enc_materials_new(request->alloc, request->requested_alg);
     if (!enc_mat) goto err;
--- a/source/default_cmm.c
+++ b/source/default_cmm.c
@@ -185,3 +185,10 @@ int aws_cryptosdk_default_cmm_set_alg_id(struct aws_cryptosdk_cmm *cmm, enum aws
     self->default_alg = alg_id;
     return AWS_OP_SUCCESS;
 }
+
+bool aws_cryptosdk_default_cmm_is_valid(const struct aws_cryptosdk_cmm *cmm) {
+    AWS_PRECONDITION(cmm != NULL);
+    struct default_cmm *self = (struct default_cmm *)cmm;
+    return aws_cryptosdk_cmm_base_is_valid(&self->base) && aws_allocator_is_valid(self->alloc) &&
+           aws_cryptosdk_keyring_is_valid(self->kr);
+}
--- a/verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
+++ b/verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
@@ -30,21 +30,14 @@ int on_encrypt(
     enum aws_cryptosdk_alg_id alg);
 
 void default_cmm_generate_enc_materials_harness() {
-    /* Nondet input required to init cmm */
-    struct aws_cryptosdk_keyring *keyring        = malloc(sizeof(*keyring));
-    const struct aws_cryptosdk_keyring_vt vtable = { .vt_size    = sizeof(struct aws_cryptosdk_keyring_vt),
+    const struct aws_cryptosdk_keyring_vt vtable = { .vt_size    = nondet_size_t(),
                                                      .name       = ensure_c_str_is_allocated(SIZE_MAX),
                                                      .destroy    = nondet_voidp(),
                                                      .on_encrypt = nondet_bool() ? NULL : on_encrypt,
                                                      .on_decrypt = nondet_voidp() };
-    /* Assumptions required to init cmm */
-    ensure_cryptosdk_keyring_has_allocated_members(keyring, &vtable);
-    __CPROVER_assume(aws_cryptosdk_keyring_is_valid(keyring));
-    __CPROVER_assume(keyring->vtable != NULL);
-
     /* Nondet input */
-    struct aws_cryptosdk_cmm *cmm               = ensure_default_cmm_attempt_allocation(keyring);
-    struct aws_cryptosdk_enc_materials **output = malloc(sizeof(*output));
+    struct aws_cryptosdk_cmm *cmm               = ensure_default_cmm_attempt_allocation(&vtable);
+    struct aws_cryptosdk_enc_materials **output = ensure_enc_materials_attempt_allocation();
     struct aws_cryptosdk_enc_request *request   = ensure_enc_request_attempt_allocation(MAX_TABLE_SIZE);
 
     /* Assumptions */
--- a/verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
+++ b/verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
@@ -27,40 +27,7 @@ int on_encrypt(
     struct aws_array_list *keyring_trace,
     struct aws_array_list *edks,
     const struct aws_hash_table *enc_ctx,
-    enum aws_cryptosdk_alg_id alg) {
-    /* Check validity for all inputs to avoid memory-safety violations. */
-    assert(aws_cryptosdk_keyring_is_valid(keyring));
-    assert(aws_allocator_is_valid(request_alloc));
-    assert(aws_byte_buf_is_valid(unencrypted_data_key));
-    assert(aws_cryptosdk_keyring_trace_is_valid(keyring_trace));
-    assert(aws_cryptosdk_edk_list_is_valid(edks));
-    assert(aws_cryptosdk_edk_list_elements_are_valid(edks));
-    assert((enc_ctx == NULL) || aws_hash_table_is_valid(enc_ctx));
-
-    if (unencrypted_data_key->buffer == NULL) {
-        const struct aws_cryptosdk_alg_properties *props = aws_cryptosdk_alg_props(alg);
-        if (props != NULL) {
-            __CPROVER_assume(aws_byte_buf_is_bounded(unencrypted_data_key, props->data_key_len));
-            ensure_byte_buf_has_allocated_buffer_member(unencrypted_data_key);
-            /*
-             * This satisfies the post-condition that if this keyring
-             * generated data key, it must be the right length.
-             * The nondeterminism increases coverage.
-             */
-            if (nondet_bool()) unencrypted_data_key->len = props->data_key_len;
-            unencrypted_data_key->allocator = request_alloc;
-            __CPROVER_assume(aws_byte_buf_is_valid(unencrypted_data_key));
-        }
-    } else {
-        /*
-         * If the buffer is not NULL, the byte buffer must not have been modified;
-         * however, this allocation modifies the buffer and increases coverage.
-         */
-        if (nondet_bool()) ensure_byte_buf_has_allocated_buffer_member(unencrypted_data_key);
-    }
-    int ret;
-    return ret;
-}
+    enum aws_cryptosdk_alg_id alg);
 
 void default_cmm_generate_enc_materials_harness() {
     /* Nondet input required to init cmm */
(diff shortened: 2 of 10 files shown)


verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,116 @@
...
+    __CPROVER_assume(request->enc_ctx != NULL);
+    ensure_allocated_hash_table(request->enc_ctx, MAX_NUM_ITEMS);
+    __CPROVER_assume(aws_cryptosdk_enc_request_is_valid(request));
+
+    enum aws_cryptosdk_alg_id alg_id;
+    if (nondet_bool()) request->requested_alg = alg_id;
+    // We can either add this OR add "if (!props) goto err;" after line 61 in the source code
+    // I think adding the check to the source code is better in this case.
+    // struct aws_cryptosdk_alg_properties *alg_props = aws_cryptosdk_alg_props(alg_id);
+    // __CPROVER_assume(aws_cryptosdk_alg_properties_is_valid(alg_props));
```
alex-chew, line 101: I agree that the check in the function implementation is better.

```suggestion
    // The function ensures that `aws_cryptosdk_alg_props(alg_id)` is valid.
```

alex-chew (changes requested):
LGTM, pending the comment about the `aws_cryptosdk_alg_props(alg_id)` validity assertion.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
feliperodri: I'm not really sure I understand this assumption. Is this related tot he next assumption that `cmm` has a set `alg_id`? Why do you even need the next assumption anyway?

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,116 @@
...
+
+    /* Nondet set self->default_alg to a valid alg_id */
+    __CPROVER_assume(aws_cryptosdk_default_cmm_set_alg_id(cmm, alg_id) == AWS_OP_SUCCESS);
+
+    /* Operation under verification */
+    if (__CPROVER_file_local_default_cmm_c_default_cmm_generate_enc_materials(cmm, output, request) == AWS_OP_SUCCESS) {
+        assert(aws_cryptosdk_enc_materials_is_valid(*output));
```
feliperodri, line 108: Is there any extra properties you can assert over `output`, `cmm`, or `request` in case of success or failure? @adpaco-aws @alex-chew

feliperodri (changes requested):

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
alex-chew: If successful, you can verify:

* `request->requested_alg` is a valid algorithm ID

If the call fails, you can verify:

* `request->requested_alg` is a valid algorithm ID or zero
* `output` is unchanged

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,116 @@
...
+
+    enum aws_cryptosdk_alg_id alg_id;
+    if (nondet_bool()) request->requested_alg = alg_id;
+    // We can either add this OR add "if (!props) goto err;" after line 61 in the source code
+    // I think adding the check to the source code is better in this case.
+    // struct aws_cryptosdk_alg_properties *alg_props = aws_cryptosdk_alg_props(alg_id);
+    // __CPROVER_assume(aws_cryptosdk_alg_properties_is_valid(alg_props));
+
+    /* Nondet set self->default_alg to a valid alg_id */
+    __CPROVER_assume(aws_cryptosdk_default_cmm_set_alg_id(cmm, alg_id) == AWS_OP_SUCCESS);
```
alex-chew, line 104: This misses the edge case where the CMM's `default_alg` is unset, i.e. it has the special value `DEFAULT_ALG_UNSET` (defined in `default_cmm.c`) which is not considered a valid algorithm: https://github.com/aws/aws-encryption-sdk-c/blob/d6ea33e9390960737d3c3888b014bbe229b42953/source/default_cmm.c#L176-L178

alex-chew (changes requested):

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
tegansb: I also like the `if (!props) goto err;` check in the function for this.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,116 @@
...
+        if (nondet_bool()) ensure_byte_buf_has_allocated_buffer_member(unencrypted_data_key);
+    }
+    int ret;
+    return ret;
+}
+
+void default_cmm_generate_enc_materials_harness() {
+    /* Nondet input required to init cmm */
+    struct aws_cryptosdk_keyring *keyring        = malloc(sizeof(*keyring));
+    const struct aws_cryptosdk_keyring_vt vtable = { .vt_size    = sizeof(struct aws_cryptosdk_keyring_vt),
```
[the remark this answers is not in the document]
adpaco-aws: Yes, we have it now.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,116 @@
...
+                                                     .destroy    = nondet_voidp(),
+                                                     .on_encrypt = nondet_bool() ? NULL : on_encrypt,
+                                                     .on_decrypt = nondet_voidp() };
+    /* Assumptions required to init cmm */
+    ensure_cryptosdk_keyring_has_allocated_members(keyring, &vtable);
+    __CPROVER_assume(aws_cryptosdk_keyring_is_valid(keyring));
+    __CPROVER_assume(keyring->vtable != NULL);
+
+    /* Nondet input */
+    struct aws_cryptosdk_cmm *cmm               = aws_cryptosdk_default_cmm_new(can_fail_allocator(), keyring);
```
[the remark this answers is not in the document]
adpaco-aws: I have created a new allocator function for default_cmm (which is an extension of cmm). We are not using this function anymore.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,116 @@
...
+
+    __CPROVER_assume(output != NULL);
+
+    __CPROVER_assume(request != NULL);
+    request->alloc = can_fail_allocator();
+    __CPROVER_assume(aws_allocator_is_valid(request->alloc));
+    request->enc_ctx = malloc(sizeof(*request->enc_ctx));
+    __CPROVER_assume(request->enc_ctx != NULL);
+    ensure_allocated_hash_table(request->enc_ctx, MAX_NUM_ITEMS);
+    __CPROVER_assume(aws_cryptosdk_enc_request_is_valid(request));
```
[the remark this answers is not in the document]
adpaco-aws: Added them.

adpaco-aws (commented):
Sorry, I had this PR almost updated but I wanted to inspect some stuff in detail. We have a new solution which includes as much feedback from you when possible.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
adpaco-aws: Please see the comment below:

> Sorry, there were two things wrong in here:
> * `aws_cryptsodk_algorithm_is_known` was not defined (fixed by adding header.c to Makefile).
> * Using an the `cmm` initializer instead of a nondet. allocator (which I have included now).
> So this code has been removed.

Because of this, all cases are now covered.

adpaco-aws:
@alex-chew I have tried the first one in here
> If the call fails, you can verify:
>
>    request->requested_alg is a valid algorithm ID or zero
>    output is unchanged
```
assert(!request->requested_alg || aws_cryptosdk_algorithm_is_known(request->requested_alg);
```
and
```
assert(request->requested_alg == DEFAULT_ALG_UNSET ||
           aws_cryptosdk_algorithm_is_known(request->requested_alg));
```
(in case you meant `DEFAULT_ALG_UNSET` as zero). But none of them work for me. Can you please elaborate?

Also, I am not sure how to check that `output` remains unchanged. Do we have templates for this? @feliperodri @tegansb

source/default_cmm.c
```diff
@@ -37,6 +37,10 @@ static int default_cmm_generate_enc_materials(
     struct aws_cryptosdk_cmm *cmm,
     struct aws_cryptosdk_enc_materials **output,
     struct aws_cryptosdk_enc_request *request) {
+    AWS_PRECONDITION(cmm != NULL);
+    AWS_PRECONDITION(output != NULL);
```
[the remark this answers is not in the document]
alex-chew: You are correct that `output` is an out parameter for the generated encryption materials, and at the start of the function it is typically not the case that `**output` is a valid encryption materials struct. A stronger precondition that would make sense might be to assert that `output` is writable for `sizeof(*output)` bytes.

alex-chew:
> @alex-chew I have tried the first one in here
> 
> > If the call fails, you can verify:
> > request->requested_alg is a valid algorithm ID or zero
> > output is unchanged
> 
> ```
> assert(!request->requested_alg || aws_cryptosdk_algorithm_is_known(request->requested_alg);
> ```
> 
> and
> 
> ```
> assert(request->requested_alg == DEFAULT_ALG_UNSET ||
>            aws_cryptosdk_algorithm_is_known(request->requested_alg));
> ```
> 
> (in case you meant `DEFAULT_ALG_UNSET` as zero). But none of them work for me. Can you please elaborate?

Oops, that's my bad. If the call fails, it should be the case that `request->requested_alg` is valid or *unchanged* (not zero). Sorry for the confusion. In retrospect, however, I'm not sure this is worth verifying.

> Also, I am not sure how to check that `output` remains unchanged. Do we have templates for this? @feliperodri @tegansb

You can save its value prior to calling the function under verification and check that its value is unchanged by the call. e.g. https://github.com/aws/aws-encryption-sdk-c/blob/master/verification/cbmc/proofs/aws_cryptosdk_priv_hdr_parse_alg_id/aws_cryptosdk_priv_hdr_parse_alg_id_harness.c

adpaco-aws:
@alex-chew Ok, I have removed the assert in case of failure.

I have also added code to check that output remains unchanged at the top level (i.e., it performs a cast of the struct and compares the values as an array, but does not check for changes on structures being pointed to by its members). The main reason to leave it as it is performance, since doing these checks would be time-consuming and the structures are not really used in the function under verification (remember that `output` is just a placeholder in this function).

source/default_cmm.c
```diff
@@ -180,3 +185,10 @@ int aws_cryptosdk_default_cmm_set_alg_id(struct aws_cryptosdk_cmm *cmm, enum aws
     self->default_alg = alg_id;
     return AWS_OP_SUCCESS;
 }
+
+bool aws_cryptosdk_default_cmm_is_valid(const struct aws_cryptosdk_cmm *cmm) {
+    AWS_PRECONDITION(cmm != NULL);
```
feliperodri, line 190: It should return false in case `cmm` is NULL, not fail on an assertion.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,71 @@
...
+                                                     .on_encrypt = nondet_bool() ? NULL : on_encrypt,
+                                                     .on_decrypt = nondet_voidp() };
+    /* Nondet input */
+    struct aws_cryptosdk_cmm *cmm              = ensure_default_cmm_attempt_allocation(&vtable);
+    struct aws_cryptosdk_enc_materials *output = ensure_enc_materials_attempt_allocation();
+    struct aws_cryptosdk_enc_request *request  = ensure_enc_request_attempt_allocation(MAX_TABLE_SIZE);
+
+    /* Assumptions */
+    __CPROVER_assume(cmm != NULL);
+    __CPROVER_assume(aws_cryptosdk_default_cmm_is_valid(cmm));
```
feliperodri, line 45: You wouldn't need it if `aws_cryptosdk_default_cmm_is_valid` returned false in case `cmm` is NULL.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
```diff
@@ -0,0 +1,71 @@
...
+
+    __CPROVER_assume(output != NULL);
+
+    __CPROVER_assume(request != NULL);
+    __CPROVER_assume(aws_allocator_is_valid(request->alloc));
+    __CPROVER_assume(request->enc_ctx != NULL);
+    __CPROVER_assume(aws_cryptosdk_enc_request_is_valid(request));
```
feliperodri, line 52: This looks odd to me too. Why the first three assumptions are not part of the validity check for `aws_cryptosdk_enc_request`?

verification/cbmc/sources/make_common_data_structures.c
```diff
@@ -30,6 +31,14 @@
 #include <proof_helpers/make_common_data_structures.h>
 #include <proof_helpers/proof_allocators.h>
 
+struct default_cmm {
+    struct aws_cryptosdk_cmm base;
+    struct aws_allocator *alloc;
+    struct aws_cryptosdk_keyring *kr;
+    // Invariant: this is either DEFAULT_ALG_UNSET or is a valid algorithm ID
```
feliperodri, line 38: Avoid cpp comments.
```c
/* Invariant: this is either DEFAULT_ALG_UNSET or is a valid algorithm ID. */
```

verification/cbmc/sources/make_common_data_structures.c
```diff
@@ -253,3 +262,50 @@ void ensure_nondet_allocate_cmm_vtable_members(struct aws_cryptosdk_cmm_vt *vtab
...
+}
+
+struct aws_cryptosdk_cmm *ensure_default_cmm_attempt_allocation(const struct aws_cryptosdk_keyring_vt *vtable) {
+    /* Nondet input required to init cmm */
+    struct aws_cryptosdk_keyring *keyring = malloc(sizeof(*keyring));
+
+    /* Assumptions required to init cmm */
+    ensure_cryptosdk_keyring_has_allocated_members(keyring, vtable);
+    __CPROVER_assume(aws_cryptosdk_keyring_is_valid(keyring));
+    __CPROVER_assume(keyring->vtable != NULL);
```
feliperodri, line 297: Create a GitHub issue to investigate a way to move these assumptions to the validity function and not to the allocator.

verification/cbmc/sources/make_common_data_structures.c
```diff
@@ -253,3 +262,50 @@ void ensure_nondet_allocate_cmm_vtable_members(struct aws_cryptosdk_cmm_vt *vtab
...
+
+    struct aws_cryptosdk_cmm *cmm = malloc(sizeof(struct default_cmm));
+    if (cmm) {
+        cmm->vtable = vtable;
+    }
+
+    struct default_cmm *self = NULL;
+    if (cmm) {
+        self        = (struct default_cmm *)cmm;
+        self->alloc = can_fail_allocator();
```
feliperodri, line 307: Could it be NULL? We should be as nondet as possible here.

verification/cbmc/stubs/on_encrypt_stub.c
```diff
@@ -0,0 +1,74 @@
...
+        }
+    } else {
+        /*
+         * If the buffer is not NULL, the byte buffer must not have been modified;
+         * however, this allocation modifies the buffer and increases coverage.
+         */
+        if (nondet_bool()) ensure_byte_buf_has_allocated_buffer_member(unencrypted_data_key);
+    }
+    int ret;
+    return ret;
```
feliperodri, line 73:
```c
return nondet_int();
```

feliperodri (changes requested):

source/default_cmm.c
adpaco-aws: Right, I am happy you saw this because I had my doubts about it.

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
adpaco-aws: Removed null assumption, thanks!

verification/cbmc/proofs/default_cmm_generate_enc_materials/default_cmm_generate_enc_materials_harness.c
adpaco-aws: This is probably from my early days of proof-writing. The assumptions were not needed so I removed them.

verification/cbmc/sources/make_common_data_structures.c
adpaco-aws: I copied this from the source code, so I fixed it in both places.

verification/cbmc/sources/make_common_data_structures.c
adpaco-aws: Yes, I missed that one. It can be `NULL` now.

verification/cbmc/stubs/on_encrypt_stub.c
adpaco-aws: Done as well.

verification/cbmc/sources/make_common_data_structures.c
adpaco-aws: Opened one here: #675

alex-chew (approved):
LGTM, pending CI fixes.

source/default_cmm.c
```diff
@@ -180,3 +185,11 @@ int aws_cryptosdk_default_cmm_set_alg_id(struct aws_cryptosdk_cmm *cmm, enum aws
...
     return AWS_OP_SUCCESS;
 }
+
+bool aws_cryptosdk_default_cmm_is_valid(const struct aws_cryptosdk_cmm *cmm) {
+    if (cmm == NULL) return false;
+
+    struct default_cmm *self = (struct default_cmm *)cmm;
+    return aws_cryptosdk_cmm_base_is_valid(&self->base) && aws_allocator_is_valid(self->alloc) &&
+           aws_cryptosdk_keyring_is_valid(self->kr);
+}
```
feliperodri, line 195: Why this function wasn't written like...
```c
bool default_cmm_is_valid(const struct default_cmm *cmm) {
    return (cmm != NULL) &&
           aws_cryptosdk_cmm_base_is_valid(&cmm->base) &&
           aws_allocator_is_valid(cmm->alloc) &&
           aws_cryptosdk_keyring_is_valid(cmm->kr);
}
```
Shouldn't we validate both structures separately (i.e., on validity function for each data structure)?

verification/cbmc/sources/make_common_data_structures.c
```diff
@@ -369,4 +379,51 @@ struct aws_cryptosdk_dec_materials *dec_materials_setup(
...
+        cmm->vtable = vtable;
+    }
+
+    struct default_cmm *self = NULL;
+    if (cmm) {
+        self        = (struct default_cmm *)cmm;
+        self->alloc = nondet_bool() ? NULL : can_fail_allocator();
+        self->kr    = keyring;
+    }
+    return (struct aws_cryptosdk_cmm *)self;
```
feliperodri, line 428: Why?

verification/cbmc/stubs/on_encrypt_stub.c
```diff
@@ -0,0 +1,73 @@
...
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdlib.h>
```
feliperodri, line 16: Why do you need that include?

feliperodri (approved):
I have a few more questions, but no blockers.

source/default_cmm.c
adpaco-aws: In general, extensions of `aws_cryptosdk_cmm` are always handled like this in the source code: The type is used as a parameter/return type and the struct is cast before doing any operations on it. This way, we are following API guidelines.

I am not sure what you mean by "validate both structures separately". We have `aws_cryptosdk_cmm_base_is_valid` for validating `aws_cryptosdk_cmm` structs and this one to validate `default_cmm` structs.

verification/cbmc/stubs/on_encrypt_stub.c
adpaco-aws: You are right, this was not required.

PR #647, opened 2020-12-02
Adds CBMC proofs for aws_cryptosdk_priv_hdr_parse_{frame,iv}_len 

tegansb:
Adds harness, Makefile and yaml for aws_cryptosdk_priv_hdr_parse_frame_len and aws_cryptosdk_priv_hdr_parse_iv_len.

Proofs are for split version of hdr_parse in #641.
Uses ensures, is bounded and validity functions from #632.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #638, opened 2020-11-23
Add CBMC proofs for various functions in cipher.c

adpaco-aws:
This PR adds proofs for the following functions in cipher.c:
* aws_cryptosdk_private_algorithm_message_id_len
* aws_cryptosdk_private_commitment_eq
* aws_cryptosdk_private_derive_key
* aws_cryptosdk_private_derive_key_v1
* aws_cryptosdk_private_derive_key_v2

It also updates `ensure_alg_properties_attempt_allocation` to initialize pointers within alg_properties structures, adds two new utils function to compare byte_bufs and content/data keys.

Finally, it updates existing proofs that were failing due to changes introduced in nondet. allocation:
* aws_cryptosdk_serialize_frame

Proofs not breaking because of these changes should be updated regardless to follow the new allocation style for alg_properties. Issue created here: #660 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #608, opened 2020-10-14
Update to aws_cryptosdk_enc_ctx_size proof

tegansb:
*Description of changes:*

Added non-NULL assumptions to harness of proof for enc_ctx_size. Updated the generator for the hash with assumptions that the key and value generated are non-NULL. 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #657, opened 2020-12-07
Adds CBMC proof for aws_cryptosdk_hdr_write

tegansb:
Adds harness, Makefile and yaml for aws_cryptosdk_priv_hdr_write. 

# Check any applicable:


verification/cbmc/proofs/aws_cryptosdk_hdr_write/Makefile
```diff
@@ -0,0 +1,85 @@
...
+
+# Remove function bodies to deal with CBMC function pointer removal
+# All functions removed are those removed in aws_cryptosdk_enc_ctx_serialize,
+# which is called from aws_cryptosdk_hdr_write. 
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_array_list_sort
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_byte_buf_write
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_cryptosdk_enc_ctx_size
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_cryptosdk_hash_elems_array_init
+
+UNWINDSET += array_list_item_generator.0:$(shell echo $$((1 + $(MAX_TABLE_SIZE))))
```
feliperodri, line 78: `$(call addone,$(MAX_TABLE_SIZE))`

verification/cbmc/proofs/aws_cryptosdk_hdr_write/Makefile
```diff
@@ -0,0 +1,85 @@
...
+UNWINDSET += aws_cryptosdk_edk_list_elements_are_bounded.0:$(call addone,$(MAX_EDK_LIST_ITEMS))
+UNWINDSET += aws_cryptosdk_edk_list_elements_are_valid.0:$(call addone,$(MAX_EDK_LIST_ITEMS))
+UNWINDSET += aws_cryptosdk_enc_ctx_serialize.0:$(shell echo $$((1 + $(MAX_TABLE_SIZE))))
```
feliperodri, line 81: `$(call addone,$(MAX_TABLE_SIZE))`

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,94 @@
...
+
+#include <aws/cryptosdk/header.h>
+#include <aws/cryptosdk/private/header.h>
+#include <make_common_data_structures.h>
+#include <proof_helpers/make_common_data_structures.h>
+#include <proof_helpers/proof_allocators.h>
+#include <proof_helpers/utils.h>
+
+// A generator function as described in the comment in aws_cryptosdk_hash_elems_array_init_stub.c.
+// Also see line 33 of the Makefile.
```
feliperodri, line 24: Avoid c++ comments.
```c
/**
 * A generator function as described in the comment in aws_cryptosdk_hash_elems_array_init_stub.c.
 * Also see line 33 of the Makefile.
 */
```

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,94 @@
...
+void array_list_item_generator(struct aws_array_list *elems) {
+    assert(elems->item_size == sizeof(struct aws_hash_element));
+    for (size_t index = 0; index < elems->length; ++index) {
+        struct aws_hash_element *val = (struct aws_hash_element *)((uint8_t *)elems->data + (elems->item_size * index));
+        // Due to the checks in aws_cryptosdk_enc_ctx_size, no string can have a length > UINT16_MAX
```
feliperodri, line 29:
```c
/* Due to the checks in aws_cryptosdk_enc_ctx_size, no string can have a length > UINT16_MAX. */
```

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,94 @@
...
+        struct aws_string *key = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(aws_string_is_valid(key));
+        __CPROVER_assume(key->len <= UINT16_MAX);
+        val->key                 = key;
```
feliperodri, line 33: Indentation.

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,94 @@
...
+void aws_cryptosdk_hdr_write_harness() {
+    /* Nondet Input */
+    struct aws_cryptosdk_hdr *hdr = ensure_nondet_hdr_has_allocated_members(MAX_TABLE_SIZE);
+    size_t *bytes_written;
+    uint8_t *outbuf;
+    size_t outlen;
+
+    /* Assumptions */
+    __CPROVER_assume(aws_cryptosdk_hdr_members_are_bounded(hdr, MAX_EDK_LIST_ITEMS, MAX_BUFFER_SIZE));
+    __CPROVER_assume(IMPLIES(hdr != NULL, aws_byte_buf_is_bounded(&hdr->iv, MAX_IV_LEN)));
```
feliperodri, line 50: Nice! I love this style.

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,94 @@
...
+
+    /* Precondition: The edk list has allocated list elements */
+    ensure_cryptosdk_edk_list_has_allocated_list_elements(&hdr->edk_list);
+    __CPROVER_assume(aws_cryptosdk_hdr_is_valid(hdr));
+
+    __CPROVER_assume(hdr->enc_ctx.p_impl != NULL);
+    ensure_hash_table_has_valid_destroy_functions(&hdr->enc_ctx);
+    size_t empty_slot_idx;
+    __CPROVER_assume(aws_hash_table_has_an_empty_slot(&hdr->enc_ctx, &empty_slot_idx));
```
feliperodri, line 59: Why do you need this assumption and why it's not mapped to an assertion in the actual code? Could you add a comment?

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,94 @@
...
+
+    ASSUME_VALID_MEMORY_COUNT(outbuf, outlen);
+    ASSUME_VALID_MEMORY(bytes_written);
```
feliperodri, line 62: Awesome!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,94 @@
...
+    /* Operation under verification */
+    aws_cryptosdk_hdr_write(hdr, bytes_written, outbuf, outlen);
+
+    /* Assertions */
+    assert(aws_cryptosdk_hdr_is_valid(hdr));
+    assert_byte_buf_equivalence(&hdr->iv, &old_iv, &old_byte_from_iv);
+    assert_byte_buf_equivalence(&hdr->auth_tag, &old_auth_tag, &old_byte_from_auth_tag);
+    assert_byte_buf_equivalence(&hdr->message_id, &old_message_id, &old_byte_from_message_id);
+    assert_byte_buf_equivalence(&hdr->alg_suite_data, &old_alg_suite_data, &old_byte_from_alg_suite_data);
+    check_hash_table_unchanged(&hdr->enc_ctx, &old_enc_ctx);
```
feliperodri, line 93: Kudos @tegansb

feliperodri (approved):
LGTM. I left a few comments to improve documentation.

alex-chew (commented):
Overall LGTM. I'm curious about the `__CPROVER_assume(aws_hash_table_has_an_empty_slot(&hdr->enc_ctx, &empty_slot_idx));` assumption that Felipe commented on as well. I also noticed that one of the `t_header.c` tests in the ubuntu-x64 build - have you checked whether it's another case of incorrectly-initialized test data?

tegansb:
> Overall LGTM. I'm curious about the `__CPROVER_assume(aws_hash_table_has_an_empty_slot(&hdr->enc_ctx, &empty_slot_idx));` assumption that Felipe commented on as well. I also noticed that one of the `t_header.c` tests in the ubuntu-x64 build - have you checked whether it's another case of incorrectly-initialized test data?

Yes, I've looked into it and it's another case of incorrectly-initialized test data! It's fixed now and the tests are passing. I'm adding a comment about the empty slot assumption!

tegansb:
@alex-chew @feliperodri I just noticed that the proof for hdr_write didn't run in CI, even though I included the yaml in this PR? Any reason you can think of for that?

verification/cbmc/proofs/aws_cryptosdk_hdr_write/Makefile
```diff
@@ -0,0 +1,86 @@
...
+PROOF_SOURCES += $(PROOF_STUB)/aws_default_allocator_stub.c
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE)
+
+# Remove function bodies to deal with CBMC function pointer removal
+# All functions removed are those removed in aws_cryptosdk_enc_ctx_serialize,
+# which is called from aws_cryptosdk_hdr_write. 
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_array_list_sort
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_byte_buf_write
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_cryptosdk_enc_ctx_size
+REMOVE_FUNCTION_BODIES += --remove-function-body aws_cryptosdk_hash_elems_array_init
```
adpaco-aws, line 77: Use `REMOVE_FUNCTION_BODY` and remove `--remove-function-body`, as in #647

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
```diff
@@ -0,0 +1,98 @@
...
+    assert(elems->item_size == sizeof(struct aws_hash_element));
+    for (size_t index = 0; index < elems->length; ++index) {
+        struct aws_hash_element *val = (struct aws_hash_element *)((uint8_t *)elems->data + (elems->item_size * index));
+        /* Due to the checks in aws_cryptosdk_enc_ctx_size, no string can have a length > UINT16_MAX. */
+        struct aws_string *key = ensure_string_is_allocated_nondet_length();
+        __CPROVER_assume(aws_string_is_valid(key));
+        /* Due to the cast to uint16, the entire size of the enc_ctx must be less than < UINT16_MAX
+         * This is a simple way to ensure this without a call to enc_ctx_size. */
+        __CPROVER_assume(key->len <= UINT8_MAX);
+        val->key                 = key;
```
adpaco-aws, line 37: Maybe this code can be moved to sources/utils.c if #638 is merged before? Then we would have the harness function alone in this file.

adpaco-aws (commented):
The proof seems to be having similar problems to the ones in #647 but other than that it looks great!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/Makefile
tegansb: Updated!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/Makefile
tegansb: Updated!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/Makefile
tegansb: Looks like I actually don't even need to remove these functions, as the stubs get used automatically so I just removed this!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: I'm open to moving this function depending on what everyone would prefer @feliperodri @alex-chew. The reason this function is in the harness for both this and the hdr_parse proofs is because the type of data structure generated by array_list_generator can change depending on the proof. So the general `aws_cryptosdk_hash_elems_array_init_stub.c `lives in the stub directory but the `array_list_generator` that generates the array elements lives in the harness itself. In the proofs I wrote, we're always generating elements of the enc_ctx which is why they look so similar. Even within the proofs I wrote, there is a small difference about whether the length is bounded by UINT8_MAX and UINT16_MAX. Nevertheless, so far we're really only used this array_list_item_generator in a similar way to the above, so we could move it to a `array_list_item_generator_enc_ctx_stub` or something of the kind to reuse it and make sure it's clearly documented.

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: Updated!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: Updated!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: So I've tried changing the indentation so it's like `val->value = value;` a few lines down but clang-format seems to insist on this style?

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: Thanks!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: Great point! So I included this assumption as it is used in the `enc_ctx_serialize` proof and `hdr_write` calls that function. However, I've just experimented locally and it seems this assumption actually isn't needed (and could also be removed from the `enc_ctx_serialize` proof). I believe this assumption in general is used to guarantee that the loop unwinding values given to CBMC are sufficient when we are iterating over a hash_table with an aws_hash_iter. Since `enc_ctx_size`(where such a loop occurs) is stubbed out in both this proof and `enc_ctx_serialize`, this assumption isn't actually needed.
We had a discussion a while back about mapping this assumption to the source code and DSN gave a good explanation of what this function does and why we don't map it directly to the actual code (https://github.com/aws/aws-encryption-sdk-c/pull/608#discussion_r508500101).

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: Thank you!

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: Thanks :)

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
tegansb: Actually I think I was wrong above. I've just removed the assumptions from enc_ctx_size and the proof remains unchanged. Perhaps this assumption was once needed to guarantee termination but no longer?

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
adpaco-aws: That makes sense to me. I agree this subtle distinction between generators needs to be well documented.

verification/cbmc/proofs/aws_cryptosdk_hdr_write/aws_cryptosdk_hdr_write_harness.c
alex-chew: Whatever clang-format accepts/does is correct, and CI enforces this.

alex-chew (approved):
LGTM assuming CI passes as expected.

PR #659, opened 2020-12-08
Adds CBMC proof for sign_header

tegansb:
Adds harness, Makefile and yaml for sign_header. 

# Check any applicable:

### DIFF
--- a/source/session_encrypt.c
+++ b/source/session_encrypt.c
@@ -171,6 +171,8 @@ static int sign_header(struct aws_cryptosdk_session *session) {
     AWS_PRECONDITION(session != NULL);
     AWS_PRECONDITION(aws_allocator_is_valid(session->alloc));
     AWS_PRECONDITION(aws_cryptosdk_hdr_is_valid(&session->header));
+    AWS_PRECONDITION(session->header.iv.len <= session->alg_props->iv_len);
+    AWS_PRECONDITION(session->header.auth_tag.len <= session->alg_props->tag_len);
     AWS_PRECONDITION(aws_cryptosdk_sig_ctx_is_valid(session->signctx));
     AWS_PRECONDITION(session->state == ST_GEN_KEY);
     AWS_PRECONDITION(session->mode == AWS_CRYPTOSDK_ENCRYPT);
--- a/source/session_encrypt.c
+++ b/source/session_encrypt.c
@@ -168,12 +168,10 @@ static int build_header(struct aws_cryptosdk_session *session, struct aws_crypto
 }
 
 static int sign_header(struct aws_cryptosdk_session *session) {
-    AWS_PRECONDITION(session != NULL);
-    AWS_PRECONDITION(aws_allocator_is_valid(session->alloc));
-    AWS_PRECONDITION(aws_cryptosdk_hdr_is_valid(&session->header));
+    AWS_PRECONDITION(aws_cryptosdk_session_is_valid(session));
+    AWS_PRECONDITION(session->alg_props->impl->cipher_ctor != NULL);
     AWS_PRECONDITION(session->header.iv.len <= session->alg_props->iv_len);
     AWS_PRECONDITION(session->header.auth_tag.len <= session->alg_props->tag_len);
-    AWS_PRECONDITION(aws_cryptosdk_sig_ctx_is_valid(session->signctx));
     AWS_PRECONDITION(session->state == ST_GEN_KEY);
     AWS_PRECONDITION(session->mode == AWS_CRYPTOSDK_ENCRYPT);
     session->header_size = aws_cryptosdk_hdr_size(&session->header);
--- a/source/session_encrypt.c
+++ b/source/session_encrypt.c
@@ -191,11 +191,11 @@ static int sign_header(struct aws_cryptosdk_session *session) {
     // see what happened. It also makes sure that the header is fully initialized,
     // again just in case some bug doesn't overwrite them properly.
 
-    if (session->header.iv.len) {
+    if (session->header.iv.len != 0) {
         assert(session->header.iv.buffer);
         memset(session->header.iv.buffer, 0x42, session->header.iv.len);
     }
-    if (session->header.auth_tag.len) {
+    if (session->header.auth_tag.len != 0) {
         assert(session->header.auth_tag.buffer);
         memset(session->header.auth_tag.buffer, 0xDE, session->header.auth_tag.len);
     }
@@ -219,17 +219,17 @@ static int sign_header(struct aws_cryptosdk_session *session) {
     if (rv) return AWS_OP_ERR;
 
     if (session->alg_props->msg_format_version == AWS_CRYPTOSDK_HEADER_VERSION_1_0) {
-        if (session->header.iv.len) {
+        if (session->header.iv.len != 0) {
             assert(session->header.iv.buffer);
             memcpy(session->header.iv.buffer, authtag.buffer, session->header.iv.len);
         }
-        if (session->header.auth_tag.len) {
+        if (session->header.auth_tag.len != 0) {
             assert(session->header.auth_tag.buffer);
             memcpy(
                 session->header.auth_tag.buffer, authtag.buffer + session->header.iv.len, session->header.auth_tag.len);
         }
     } else {
-        if (session->header.auth_tag.len) {
+        if (session->header.auth_tag.len != 0) {
             assert(session->header.auth_tag.buffer);
             memcpy(session->header.auth_tag.buffer, authtag.buffer, session->header.auth_tag.len);
         }
--- a/source/session_encrypt.c
+++ b/source/session_encrypt.c
@@ -168,6 +168,12 @@ static int build_header(struct aws_cryptosdk_session *session, struct aws_crypto
 }
 
 static int sign_header(struct aws_cryptosdk_session *session) {
+    AWS_PRECONDITION(session != NULL);
+    AWS_PRECONDITION(aws_allocator_is_valid(session->alloc));
+    AWS_PRECONDITION(aws_cryptosdk_hdr_is_valid(&session->header));
+    AWS_PRECONDITION(aws_cryptosdk_sig_ctx_is_valid(session->signctx));
+    AWS_PRECONDITION(session->state == ST_GEN_KEY);
+    AWS_PRECONDITION(session->mode == AWS_CRYPTOSDK_ENCRYPT);
     session->header_size = aws_cryptosdk_hdr_size(&session->header);
 
     if (session->header_size == 0) {
@@ -183,8 +189,14 @@ static int sign_header(struct aws_cryptosdk_session *session) {
     // see what happened. It also makes sure that the header is fully initialized,
     // again just in case some bug doesn't overwrite them properly.
 
-    memset(session->header.iv.buffer, 0x42, session->header.iv.len);
-    memset(session->header.auth_tag.buffer, 0xDE, session->header.auth_tag.len);
+    if (session->header.iv.len) {
+        assert(session->header.iv.buffer);
+        memset(session->header.iv.buffer, 0x42, session->header.iv.len);
+    }
+    if (session->header.auth_tag.len) {
+        assert(session->header.auth_tag.buffer);
+        memset(session->header.auth_tag.buffer, 0xDE, session->header.auth_tag.len);
+    }
 
     size_t actual_size;
 
@@ -205,10 +217,20 @@ static int sign_header(struct aws_cryptosdk_session *session) {
     if (rv) return AWS_OP_ERR;
 
     if (session->alg_props->msg_format_version == AWS_CRYPTOSDK_HEADER_VERSION_1_0) {
-        memcpy(session->header.iv.buffer, authtag.buffer, session->header.iv.len);
-        memcpy(session->header.auth_tag.buffer, authtag.buffer + session->header.iv.len, session->header.auth_tag.len);
+        if (session->header.iv.len) {
+            assert(session->header.iv.buffer);
+            memcpy(session->header.iv.buffer, authtag.buffer, session->header.iv.len);
+        }
+        if (session->header.auth_tag.len) {
+            assert(session->header.auth_tag.buffer);
+            memcpy(
+                session->header.auth_tag.buffer, authtag.buffer + session->header.iv.len, session->header.auth_tag.len);
+        }
     } else {
-        memcpy(session->header.auth_tag.buffer, authtag.buffer, session->header.auth_tag.len);
+        if (session->header.auth_tag.len) {
+            assert(session->header.auth_tag.buffer);
+            memcpy(session->header.auth_tag.buffer, authtag.buffer, session->header.auth_tag.len);
+        }
     }
 
     // Re-serialize the header now that we know the auth tag
--- a/verification/cbmc/proofs/sign_header/Makefile
+++ b/verification/cbmc/proofs/sign_header/Makefile
@@ -76,10 +76,9 @@ PROOF_SOURCES += $(PROOF_STUB)/aws_array_list_sort_noop_stub.c
 PROOF_SOURCES += $(PROOF_STUB)/aws_cryptosdk_enc_ctx_size_stub.c
 PROOF_SOURCES += $(PROOF_STUB)/aws_cryptosdk_hash_elems_array_init_stub.c
 PROOF_SOURCES += $(PROOF_STUB)/aws_default_allocator_stub.c
+# Stubbing out aws_cryptosdk_hdr_write leads to a huge gain in performance
 PROOF_SOURCES += $(PROOF_STUB)/hdr_write_stub.c
 PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE)
-
-REMOVE_FUNCTION_BODY += aws_array_list_get_at_ptr
 
 # We abstract these functions because manual inspection demonstrates that they are unreachable,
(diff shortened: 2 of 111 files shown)


verification/cbmc/proofs/sign_header/sign_header_harness.c
```diff
@@ -0,0 +1,99 @@
...
+
+    enum aws_cryptosdk_alg_id alg_id;
+    struct aws_cryptosdk_alg_properties *props = aws_cryptosdk_alg_props(alg_id);
+    __CPROVER_assume(aws_cryptosdk_alg_properties_is_valid(props));
+    session->alg_props = props;
+
+    struct aws_cryptosdk_hdr *hdr = ensure_nondet_hdr_has_allocated_members(MAX_TABLE_SIZE);
+    __CPROVER_assume(aws_cryptosdk_hdr_members_are_bounded(hdr, MAX_EDK_LIST_ITEMS, MAX_BUFFER_SIZE));
+    __CPROVER_assume(IMPLIES(hdr != NULL, aws_byte_buf_is_bounded(&hdr->iv, session->alg_props->iv_len)));
+    __CPROVER_assume(IMPLIES(hdr != NULL, aws_byte_buf_is_bounded(&hdr->auth_tag, session->alg_props->tag_len)));
```
alex-chew, line 60: Without being familiar with `aws_cryptosdk_hdr_members_are_bounded`, these lines are confusing to me - if line 58 assumes that the all header members are bounded, why are lines 59-60 necessary to assume that some particular header members are bounded?

verification/cbmc/proofs/sign_header/sign_header_harness.c
```diff
@@ -0,0 +1,99 @@
...
+    ensure_cryptosdk_edk_list_has_allocated_list_elements(&hdr->edk_list);
+    __CPROVER_assume(aws_cryptosdk_hdr_is_valid(hdr));
+
+    __CPROVER_assume(hdr->enc_ctx.p_impl != NULL);
+    ensure_hash_table_has_valid_destroy_functions(&hdr->enc_ctx);
+    size_t empty_slot_idx;
+    __CPROVER_assume(aws_hash_table_has_an_empty_slot(&hdr->enc_ctx, &empty_slot_idx));
```
alex-chew, line 67: As in #657 it would be nice to explain the purpose of this assumption.

alex-chew (approved):
Mostly LGTM, just a few nitpicks about naming/comments.

verification/cbmc/proofs/sign_header/sign_header_harness.c
```diff
@@ -0,0 +1,99 @@
...
+#include <aws/cryptosdk/private/cipher.h>
+#include <aws/cryptosdk/private/session.h>
+#include <aws/cryptosdk/session.h>
+#include <cbmc_invariants.h>
+#include <make_common_data_structures.h>
+#include <proof_helpers/make_common_data_structures.h>
+
+/**
+ * A generator function as described in the comment in aws_cryptosdk_hash_elems_array_init_stub.c.
+ * Also see line 33 of the Makefile.
```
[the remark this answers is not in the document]
tegansb: Updated!

verification/cbmc/proofs/sign_header/sign_header_harness.c
tegansb: So this has been partially moved into hdr_setup, but basically the generic header setup assumes that all the byte_bufs fields of hdr are bounded by a MAX_BUFFER_SIZE. This MAX_BUFFER_SIZE comes from Makefile.aws_byte_buf and is set to 1028. However, for the iv and auth_tag having a buffer that large will actually result in overflow. Therefore I add the additional assumptions that they are bounded by session->alg_props->iv_len and session->alg_props->tag_len respectively (in fact I think they should be equal?). Is that what you would expect for the iv and auth_tag buffers?

source/session_encrypt.c
```diff
@@ -168,6 +168,14 @@ static int build_header(struct aws_cryptosdk_session *session, struct aws_crypto
...
 
 static int sign_header(struct aws_cryptosdk_session *session) {
+    AWS_PRECONDITION(session != NULL);
+    AWS_PRECONDITION(aws_allocator_is_valid(session->alloc));
+    AWS_PRECONDITION(aws_cryptosdk_hdr_is_valid(&session->header));
+    AWS_PRECONDITION(session->header.iv.len <= session->alg_props->iv_len);
+    AWS_PRECONDITION(session->header.auth_tag.len <= session->alg_props->tag_len);
+    AWS_PRECONDITION(aws_cryptosdk_sig_ctx_is_valid(session->signctx));
+    AWS_PRECONDITION(session->state == ST_GEN_KEY);
+    AWS_PRECONDITION(session->mode == AWS_CRYPTOSDK_ENCRYPT);
```
feliperodri, line 178: We don't need an unique allocator for `session` to cope with proof complexity, but we should have a validity function for it. Could you group the general checks here into a validity function for `aws_cryptosdk_session` structures?

source/session_encrypt.c
```diff
@@ -183,8 +191,14 @@ static int sign_header(struct aws_cryptosdk_session *session) {
...
-    memset(session->header.iv.buffer, 0x42, session->header.iv.len);
-    memset(session->header.auth_tag.buffer, 0xDE, session->header.auth_tag.len);
+    if (session->header.iv.len) {
+        assert(session->header.iv.buffer);
+        memset(session->header.iv.buffer, 0x42, session->header.iv.len);
+    }
+    if (session->header.auth_tag.len) {
+        assert(session->header.auth_tag.buffer);
+        memset(session->header.auth_tag.buffer, 0xDE, session->header.auth_tag.len);
+    }
```
feliperodri, line 201: Should we assert or return error? @alex-chew

source/session_encrypt.c
```diff
@@ -205,10 +217,20 @@ static int sign_header(struct aws_cryptosdk_session *session) {
...
+            assert(session->header.auth_tag.buffer);
+            memcpy(
+                session->header.auth_tag.buffer, authtag.buffer + session->header.iv.len, session->header.auth_tag.len);
+        }
     } else {
-        memcpy(session->header.auth_tag.buffer, authtag.buffer, session->header.auth_tag.len);
+        if (session->header.auth_tag.len) {
+            assert(session->header.auth_tag.buffer);
+            memcpy(session->header.auth_tag.buffer, authtag.buffer, session->header.auth_tag.len);
+        }
```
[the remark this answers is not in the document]
feliperodri: +1

source/session_encrypt.c
```diff
@@ -205,10 +219,20 @@ static int sign_header(struct aws_cryptosdk_session *session) {
     if (rv) return AWS_OP_ERR;
 
     if (session->alg_props->msg_format_version == AWS_CRYPTOSDK_HEADER_VERSION_1_0) {
-        memcpy(session->header.iv.buffer, authtag.buffer, session->header.iv.len);
-        memcpy(session->header.auth_tag.buffer, authtag.buffer + session->header.iv.len, session->header.auth_tag.len);
+        if (session->header.iv.len) {
```
feliperodri, line 222: nit: I'd also make it explicit `if (session->header.iv.len != 0)`

verification/cbmc/proofs/sign_header/Makefile
```diff
@@ -0,0 +1,104 @@
...
+DEFINES += -DMAX_IV_LEN=$(MAX_IV_LEN)
+DEFINES += -DAWS_CRYPTOSDK_HASH_ELEMS_ARRAY_INIT_GENERATOR=array_list_item_generator
+
+# The maximum number of objects is set to 2^8 = 256. This option changes that to 2^9. 
+# Without this flag, running this proof causes the CBMC error of 
+# "too many addressed objects: maximum number of objects is set to 2^n=256 (with n=8);
+# use the `--object-bits n` option to increase the maximum number"
+# The proof hits the error when the line "rv = aws_cryptosdk_hdr_write(&session->header,
+# &actual_size, session->header_copy, session->header_size);" is reached.
+CBMCFLAGS += --object-bits 9
```
feliperodri, line 45: This is wrong. You should use
```Makefile
CBMC_OBJECT_BITS = 9
```

feliperodri (changes requested):

source/session_encrypt.c
alex-chew: Short answer: assert. The condition should always be true, and if not then it's a programming error rather than a user/data error.

Long answer: I think we can actually leave out the assert, since the `aws_cryptosdk_hdr_is_valid(&session->header)` precondition implies that the `session->header.iv` and `session->header.auth_tag` are `aws_byte_buf_is_valid`, which in turn implies that `iv.len` implies `iv.buffer` and likewise for `auth_tag`. But the assert is valuable as documentation of this intent.

verification/cbmc/proofs/sign_header/sign_header_harness.c
alex-chew: I see, I think you're right - the buffer lengths are indeed equal-to/bound-by the `alg_props->{iv,tag}_len` values. I had missed the different bounds in my original comment.

source/session_encrypt.c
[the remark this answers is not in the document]
alex-chew: Same answer as https://github.com/aws/aws-encryption-sdk-c/pull/659#discussion_r546879419; assert makes more sense than an error, but even the assert serves as documentation and is itself not necessary

source/session_encrypt.c
alex-chew: Digging deeper, it looks like the only callsite of `sign_header` appears after a successfull call to `build_header`, which initializes these buffers to have positive length (since all supported algorithm suites use 12- and 16-byte IVs and authentication tag lengths, respectively). So we can change the guards/asserts to preconditions (i.e. `byte_buf_is_valid` and `buf.len > 0` for each of the IV and auth tag).

source/session_encrypt.c
tegansb: Updated to make explicit for all cases.

verification/cbmc/proofs/sign_header/Makefile
tegansb: Updated!

source/session_encrypt.c
```diff
@@ -168,6 +168,12 @@ static int build_header(struct aws_cryptosdk_session *session, struct aws_crypto
 }
 
 static int sign_header(struct aws_cryptosdk_session *session) {
+    AWS_PRECONDITION(session != NULL);
+    AWS_PRECONDITION(aws_allocator_is_valid(session->alloc));
+    AWS_PRECONDITION(aws_cryptosdk_hdr_is_valid(&session->header));
+    AWS_PRECONDITION(aws_cryptosdk_sig_ctx_is_valid(session->signctx));
+    AWS_PRECONDITION(session->state == ST_GEN_KEY);
+    AWS_PRECONDITION(session->mode == AWS_CRYPTOSDK_ENCRYPT);
```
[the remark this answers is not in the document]
tegansb: Updated!

source/session_encrypt.c
tegansb: Yes! This has been updated!

source/session_encrypt.c
```diff
@@ -168,12 +168,10 @@ static int build_header(struct aws_cryptosdk_session *session, struct aws_crypto
...
 
 static int sign_header(struct aws_cryptosdk_session *session) {
-    AWS_PRECONDITION(session != NULL);
-    AWS_PRECONDITION(aws_allocator_is_valid(session->alloc));
-    AWS_PRECONDITION(aws_cryptosdk_hdr_is_valid(&session->header));
+    AWS_PRECONDITION(aws_cryptosdk_session_is_valid(session));
+    AWS_PRECONDITION(session->alg_props->impl->cipher_ctor != NULL);
     AWS_PRECONDITION(session->header.iv.len <= session->alg_props->iv_len);
     AWS_PRECONDITION(session->header.auth_tag.len <= session->alg_props->tag_len);
-    AWS_PRECONDITION(aws_cryptosdk_sig_ctx_is_valid(session->signctx));
```
alex-chew: Not a blocker, but I'm curious why this is no longer needed?

alex-chew (approved):

source/session_encrypt.c
adpaco-aws: The new session validator checks that `session->signctx` is either valid or `NULL`. The function does actually check before using it on `aws_cryptosdk_sig_update` so it was probably not required before and by using the session validator we are considering more cases.

feliperodri (approved):
LGTM.

source/session_encrypt.c
adpaco-aws: @alex-chew we have investigated this issue and we have decided to remove the assumption from the harness. Note that the session validator allows `session->signctx` to be either NULL or valid. To do so, we modified a precondition in `EVP_DigestUpdate` which we think is too restrictive.

Moreover, we have opened an issue (#685) to remove this type of validators from the code base, since they cannot be mirrored as preconditions.

adpaco-aws:
The last changes include a stub for `aws_cryptosdk_hdr_write` which causes the proof to run much faster than earlier. 

@feliperodri can you take a look at this stub and let us know what you think? The original function can be found here: https://github.com/aws/aws-encryption-sdk-c/blob/6a2ee0e0a3d7857aa138244f3b944be7d9b7d9b2/source/header.c#L472-L568

verification/cbmc/stubs/hdr_write_stub.c
```diff
@@ -0,0 +1,46 @@
...
+#include <proof_helpers/make_common_data_structures.h>
+
+/*
+ * Stub for aws_cryptosdk_hdr_write
+ * The original function declares an aws_byte_buf output variable where
+ * different components of the header get written, but the actual output
+ * of the function is bytes_written. Therefore, we assign a nondet. value
+ * to this variable in case of success or zero both outbuf and outlen in
+ * case of failure
+ */
```
feliperodri, line 27: nit: Suggestion:

```c
/**
 * The original aws_cryptosdk_hdr_write function declares an aws_byte_buf
 * output variable where different components of the header get written,
 * however, the actual output of the function is bytes_written. Therefore,
 * we assign a nondeterministic value to bytes_written in case of success
 * or zero both outbuf and outlen in case of failure.
 */
```

feliperodri: The stub looks great. How much gain in performance?

feliperodri (approved):
LGTM, modulo a few suggestions.

verification/cbmc/proofs/sign_header/Makefile
```diff
@@ -0,0 +1,102 @@
...
+PROJECT_SOURCES += $(SRCDIR)/source/session.c
+PROJECT_SOURCES += $(SRCDIR)/source/session_encrypt.c
+PROJECT_SOURCES += $(SRCDIR)/source/utils.c
+
+PROOF_SOURCES += $(COMMON_PROOF_SOURCE)/make_common_data_structures.c
+PROOF_SOURCES += $(COMMON_PROOF_SOURCE)/proof_allocators.c
+PROOF_SOURCES += $(COMMON_PROOF_SOURCE)/utils.c
+PROOF_SOURCES += $(COMMON_PROOF_STUB)/aws_array_list_defined_type.c
+PROOF_SOURCES += $(COMMON_PROOF_STUB)/aws_byte_buf_write_stub.c
+PROOF_SOURCES += $(COMMON_PROOF_STUB)/aws_hash_iter_overrides.c
```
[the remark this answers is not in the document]
adpaco-aws: Done! #686

verification/cbmc/proofs/sign_header/Makefile
```diff
@@ -0,0 +1,102 @@
...
+PROOF_SOURCES += $(COMMON_PROOF_STUB)/memcmp_override.c
+PROOF_SOURCES += $(COMMON_PROOF_STUB)/memcpy_override_havoc.c
+PROOF_SOURCES += $(COMMON_PROOF_STUB)/memset_override_havoc.c
+PROOF_SOURCES += $(PROOF_SOURCE)/cbmc_invariants.c
+PROOF_SOURCES += $(PROOF_SOURCE)/make_common_data_structures.c
+PROOF_SOURCES += $(PROOF_STUB)/aws_array_list_sort_noop_stub.c
+PROOF_SOURCES += $(PROOF_STUB)/aws_cryptosdk_enc_ctx_size_stub.c
+PROOF_SOURCES += $(PROOF_STUB)/aws_cryptosdk_hash_elems_array_init_stub.c
+PROOF_SOURCES += $(PROOF_STUB)/aws_default_allocator_stub.c
+PROOF_SOURCES += $(PROOF_STUB)/hdr_write_stub.c
```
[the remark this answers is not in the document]
adpaco-aws: Done!

verification/cbmc/proofs/sign_header/Makefile
```diff
@@ -0,0 +1,102 @@
...
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE)
+
+REMOVE_FUNCTION_BODY += aws_array_list_get_at_ptr
```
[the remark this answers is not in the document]
adpaco-aws: Done!

verification/cbmc/stubs/hdr_write_stub.c
adpaco-aws: Thanks! It was ~40min on my dedicated machine before, but only ~10min now 😄

Issue #672, opened 2020-12-21
CBMC proof for aws_cryptosdk_keyring_trace_copy_all function eat up all CI resources


adpaco-aws:
There is an ongoing issue on CI with CBMC proof for `aws_cryptosdk_keyring_trace_copy_all`. The solver is running out of memory so the proof fails. We are investigating how to deal with this issue.

For the time being, we are going to disable the execution of this proof. We will restore it once we find a solution to it.

adpaco-aws:
#677 solved the problem but using the release version of CBMC on CI caused this problem to reappear. We made a change in CI to take care of this, but in the future it would be better to have this proof (or the SAT solver) executing in a separate cluster.

PR #674, opened 2020-12-22
Set concurrency limit for litani jobs

adpaco-aws:
This PR adds a concurrency limit for litani run-build jobs as a solution to #672. It also restores CBMC proof `aws_cryptosdk_keyring_trace_copy_all` which was temporarily disabled in #673.

# Check any applicable:

### DIFF
--- a/verification/cbmc/proofs/run-cbmc-proofs.py
+++ b/verification/cbmc/proofs/run-cbmc-proofs.py
@@ -163,8 +163,8 @@ def run_build(litani, jobs):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
-    # Set a limit on the number of concurrent jobs based on CPUs
-    # https://github.com/aws/aws-encryption-sdk-c/pull/673
+    # Restrict concurrency during CI to avoid 'out of memory' errors
+    # when proofs that require a lot of memory run concurrently
     else:
         factor   = 0.75
         cpu_num  = os.cpu_count()
--- a/verification/cbmc/proofs/run-cbmc-proofs.py
+++ b/verification/cbmc/proofs/run-cbmc-proofs.py
@@ -84,6 +84,11 @@ def get_args():
             "metavar": "N",
             "help": "run at most N proof jobs in parallel",
     }, {
+            "flags": ["--parallel-factor"],
+            "type": float,
+            "default": 0.75,
+            "help": "run at most (F * CPUs) jobs in parallel",
+    }, {
             "flags": ["--no-standalone"],
             "action": "store_true",
             "help": "only configure proofs: do not initialize nor run",
@@ -159,17 +164,17 @@ def get_proof_dirs(proof_root, proof_list, proof_marker):
         sys.exit(1)
 
 
-def run_build(litani, jobs):
+def run_build(litani, jobs, factor):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
     # Restrict concurrency during CI to avoid 'out of memory' errors
     # when proofs that require a lot of memory run concurrently
     else:
-        factor   = 0.75
         cpu_num  = os.cpu_count()
-        lim_jobs = int(factor * cpu_num)
-        cmd.extend(["-j", str(lim_jobs)])
+        if cpu_num is not None and cpu_num > 1:
+            lim_jobs = int(factor * cpu_num)
+            cmd.extend(["-j", str(lim_jobs)])
 
     logging.debug(" ".join(cmd))
     proc = subprocess.run(cmd)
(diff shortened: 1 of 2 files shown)


verification/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -163,6 +163,13 @@ def run_build(litani, jobs):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
+    # Set a limit on the number of concurrent jobs based on CPUs
+    # https://github.com/aws/aws-encryption-sdk-c/pull/673
+    else:
+        factor   = 0.75
```
markrtuttle, line 169: I appreciate you giving a good name to this constant.  I'm uncertain about whether this should be a top-level comment or hidden here in the code.

verification/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -163,6 +163,13 @@ def run_build(litani, jobs):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
+    # Set a limit on the number of concurrent jobs based on CPUs
+    # https://github.com/aws/aws-encryption-sdk-c/pull/673
```
markrtuttle, line 167: I don't think there is a lot of value in pointing to the troublesome pull request.  I think a more helpful comment would be something like "Restrict concurrency during CI to a level that allows proofs requiring a lot of memory from consuming memory and causing other proofs to fail with 'out of memory'."

markrtuttle (changes requested):

verification/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -163,6 +163,13 @@ def run_build(litani, jobs):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
+    # Restrict concurrency during CI to avoid 'out of memory' errors
+    # when proofs that require a lot of memory run concurrently
+    else:
+        factor   = 0.75
+        cpu_num  = os.cpu_count()
+        lim_jobs = int(factor * cpu_num)
```
karkhaz, line 171: If cpu_num is 1, this will end up being 0.

verification/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -163,6 +163,13 @@ def run_build(litani, jobs):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
+    # Restrict concurrency during CI to avoid 'out of memory' errors
+    # when proofs that require a lot of memory run concurrently
+    else:
+        factor   = 0.75
```
karkhaz, line 169: I suggest to make this an argument to the script with a default value of 0.75. Then there's no need for a comment at all, instead write a help text that users will see.

markrtuttle (approved):

verification/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -163,6 +163,13 @@ def run_build(litani, jobs):
...
+        cpu_num  = os.cpu_count()
```
karkhaz, line 170: This can return `None` on some platforms, you need to check for that

karkhaz (changes requested):

verification/cbmc/proofs/run-cbmc-proofs.py
adpaco-aws: Added a check for`cpu_num` being not `None` and greater than 1.

verification/cbmc/proofs/run-cbmc-proofs.py
adpaco-aws: I have added this argument as "parallel factor" as suggested. I would like to leave the comment since there is no reason to do that in a normal situation.

verification/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -159,10 +164,15 @@ def get_proof_dirs(proof_root, proof_list, proof_marker):
...
-def run_build(litani, jobs):
+def run_build(litani, jobs, factor):
     cmd = [str(litani), "run-build"]
     if jobs:
         cmd.extend(["-j", str(jobs)])
+    else:
+        cpu_num  = os.cpu_count()
+        if cpu_num is not None and cpu_num > 1:
+            lim_jobs = int(factor * cpu_num)
+            cmd.extend(["-j", str(lim_jobs)])
```
karkhaz, line 175: I think there should be an `else` case. At least log a warning, so that we can see in the logs if we're running on a platform where `cpu_num` is None.

Also, I still think this is incorrect. If `cpu_num` is 2 and the user passes `factor` as 0.25, then `lim` will still be zero. I think you should do `lim_jobs = lim_jobs if lim_jobs else 1`.

karkhaz (changes requested):

karkhaz (approved):
Approved, although a proof is failing. Apparently 0.75 is still too large to be reliable.

adpaco-aws:
Closed in favour of #677

PR #687, opened 2021-01-15
fix(default_cmm): track unset default alg separately

alex-chew:
Previously the Default CMM stored a sentinel value to indicate an unset default CMM, but under some compiler settings this showed and error. Now we store this state separately.

# Check any applicable:


robin-aws:
Just noting here that the change looks good AFAICT, but I don't want to merge it until we address the fact that CI didn't catch what looks like a legitimate issue on at least some platforms.

ajewellamz (approved):

Issue #658, opened 2020-12-08
Remove macOS Actions CI workaround for CMake version 3.19.1


alex-chew:
The hosted GitHub Actions runners for macOS recently upgraded the default CMake version to v3.19.1, leading to build failures: https://github.com/aws/aws-encryption-sdk-c/issues/650

We implemented a workaround that involves downgrading CMake to the prior version, v3.18.5: https://github.com/aws/aws-encryption-sdk-c/pull/655

We should root cause the build failure, and determine (a) what changes to make to our build configuration so that we can use newer CMake versions, or (b) whether we need to just wait for a CMake version with the regression patched.

lizroth:
From #692 

CMake 3.19.1 seems to have a problem with `aws-checksums` configuration. This appears to be a bug fixed in CMake because the build works fine with 3.19.3 and the older versions of CMake that this change upgrades.

Unfortunately 3.19.1 is *the* version in GitHub Actions' MacOS 10.15 image right now, see https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

Attempting to use Homebrew, 3.19.3 is generally exposed but is not exposed to the GHA environment.

In any event, manually pulling 3.19.3 from CMake's repository the way we were pulling 3.18.5 does the trick.

I am leaving #658 open though because eventually it would be nice to remove this version pinning entirely.

PR #692, opened 2021-01-22
fix: Force MacOS build to CMake 3.19.3.

lizroth:
*Issue #, if available:* #658

*Description of changes:* CMake 3.19.1 seems to have a problem with `aws-checksums` configuration. This appears to be a bug fixed in CMake because the build works fine with 3.19.3 and the older versions of CMake that this change upgrades.

Unfortunately 3.19.1 is *the* version in GitHub Actions' MacOS 10.15 image right now, see https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

Attempting to use Homebrew, 3.19.3 is generally exposed but is not exposed to the GHA environment.

In any event, manually pulling 3.19.3 from CMake's repository the way we were pulling 3.18.5 does the trick.

I am leaving #658 open though because eventually it would be nice to remove this version pinning entirely.

(I'll retitle #658 if this fix is acceptable and is merged.)

# Check any applicable:

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -39,8 +39,10 @@ jobs:
       # https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md
       - name: Install CMake v3.19.3
         run: |
-          brew install cmake@3.19.3
-          brew switch cmake 3.19.3
+          brew uninstall cmake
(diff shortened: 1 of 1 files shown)


alex-chew (approved):
LGTM. It's good to know there's also a simpler way to manually install CMake (versus manually copying over the relevant files).

PR #738, opened 2021-11-23
fix(examples): adjust discovery example wording

alex-chew:

# Check any applicable:

### DIFF
--- a/examples/kms_discovery.cpp
+++ b/examples/kms_discovery.cpp
@@ -201,10 +201,9 @@ int main(int argc, char **argv) {
                                       .WithKmsClient(create_kms_client(Aws::Region::EU_CENTRAL_1))
                                       .BuildDiscovery());
 
-    /* This will only attempt to decrypt using keys owned by the specified AWS
-     * account. It will succeed if the message was encrypted with any KMS key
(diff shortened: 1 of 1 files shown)


examples/kms_discovery.cpp
```diff
@@ -201,10 +201,9 @@ int main(int argc, char **argv) {
                                       .WithKmsClient(create_kms_client(Aws::Region::EU_CENTRAL_1))
                                       .BuildDiscovery());
 
-    /* This will only attempt to decrypt using keys owned by the specified AWS
-     * account. It will succeed if the message was encrypted with any KMS key
-     * owned by the specified AWS account and for which you have Decrypt
-     * permissions.
+    /* This will only attempt to decrypt using keys in the specified AWS account.
+     * It will succeed if the message was encrypted with any KMS key in the
+     * specified AWS account and for which you have Decrypt permissions.
```
juneb, line 206:
```suggestion
    /* This will attempt to decrypt using only KMS keys in the specified AWS account.
     * It will succeed if the message was encrypted with any KMS key in the
     * specified AWS account and for which you have kms:Decrypt permission.
```

smswz (approved):

PR #716, opened 2021-06-22
Integrate aws-encryption-sdk-c with awslc on Ubuntu.

bryce-shang:
*Issue #, if available:*
CryptoAlg-778

*Description of changes:*
This PR added a new crypto lib option -- [AWS libcrypto](https://github.com/awslabs/aws-lc).

*Call out:*
* I will submit new PRs to add awslc into other CI dimensions -- MacOS GitHub workflow and Windows CodeBuild.
* For this PR testing, new batch CodeBuild -- `csdk-ubuntu-latest-x64-awslc` should be added to CI.
   * The build spec file `./codebuild/ubuntu-latest-x64/batch_awslc.yml` should be used.

*Test:*
This PR also added a new Docker image `ubuntu-latest-x64-awslc.Dockerfile`, which is mirror of `ubuntu-latest-x64.Dockerfile` but lets the build spec install `latest` awslc.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #750, opened 2022-06-06
feat: Add support for AWS-LC build in ESDK

samuel40791765:
*Issue #, if available:*
Resolves `CryptoAlg-1133`

*Description of changes:*
After meeting up and discussing with the ESDK team, we're ready to continue pushing forward with this PR. The PR has been verified to still work with ESDK as of right now, and we will continue pushing forward with the changes within it. 

1. Two additional header file references in `source/cipher_openssl.c`
`source/cipher_openssl.c` uses ASN1  and OBJ specific functions (`BN_to_ASN1_INTEGER`, `d2i_ASN1_INTEGER`, `OBJ_txt2nid` etc.). 
These functions are defined in AWS-LC’s `asn1.h` & `obj.h` header file. Adding the two header file references successfully resolves all of the missing symbols. ESDK had gotten around `asn1.h` because the existing OpenSSL header `ec.h` in the code file was indirectly calling upon `asn1.h`. `obj.h` is AWS-LC specific.

2. `OpenSSL_free`
CESDK calls `i2o_ECPublicKey` which internally allocates memory using `OPENSS_malloc`. This behavior applies to both implementation of `i2o_ECPublicKey` in OpenSSL and AWS-LC. The expected way to deal with memory allocated with `OPENSSL_malloc` is to free it with `OPENSSL_free`. However, when the `OPENSSL_free` change is applied to OpenSSL builds for ESDK, the CBMC proof for `aws_cryptosdk_sig_get_pubkey` will fail. This might be due to the CMBC proof overrides not recognizing the `OpenSSL_free` function.
The only change that needs to be done here is to change `free` to `OpenSSL_free`, to deal with memory that has been passed into `i2o_ECPublicKey`.

3. HKDF
AWS-LC does not support HKDF construction with the `EVP_PKEY*` methods in OpenSSL 1.1.1.
We're backporting to ESDK's HKDF self implementation methods `aws_cryptosdk_hkdf_extract`
and `aws_cryptosdk_hkdf_expand` when building with AWS-LC for now.
AWS-LC does have support for HKDF (`HKDF_extract` & `HKDF_expand`) and the semantics of them
 are nearly identical to ESDK's self implementation. We can look to add support if there are customer
requirements to use AWS-LC's HKDF crypto implementation when building with ESDK.

4. AWS-LC CI support in ESDK
A docker file with test scripts to support AWS-LC  has been added. The scripts aim to work similarly to ESDK's CI docker file build process (`codebuild/ubuntu-latest-x64.Dockerfile`). The CI should be able to use the same buildbatch and build specs defined under `codebuild/ubuntu-latest-x64`.

The contents of this PR were mostly taken and adjusted from https://github.com/aws/aws-encryption-sdk-c/pull/716.

# Check any applicable:


source/hkdf.c
```diff
@@ -29,7 +29,18 @@ static const EVP_MD *aws_cryptosdk_get_evp_md(enum aws_cryptosdk_sha_version whi
     }
 }
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
+// AWS-LC does not support HKDF construction with the |EVP_PKEY*| methods in OpenSSL 1.1.1.
+// We're backporting to ESDK's HKDF self implementation methods |aws_cryptosdk_hkdf_extract|
+// and |aws_cryptosdk_hkdf_expand| when building with AWS-LC for now.
+// The semantics of AWS-LC's HKDF API implementations (|HKDF_extract| & |HKDF_expand|)
+// are nearly identical to ESDK's. We can look to add support if there are customer
+// requirements to use AWS-LC's HKDF crypto implementation when building with ESDK.
```
alex-chew, line 37: Can you clarify what you mean by the semantics of AWS-LC's HKDF API implementations being "nearly identical" to the ESDK's?

codebuild/bin/install-shared-deps-awslc.sh
```diff
@@ -0,0 +1,78 @@
...
+set -euxo pipefail
+
+export AWSLC_SRC_DIR=/tmp/awslc
+export INSTALL_DIR=/deps/install
+export LD_LIBRARY_PATH=${INSTALL_DIR}
+export NUM_CPU_THREADS=$(nproc)
+
+function download_awslc() {
+    AWSLC_GIT_URL='https://github.com/awslabs/aws-lc.git'
+    AWSLC_GIT_REF='main'
```
alex-chew, line 20: We should pin a particular AWS-LC version as we do for the OpenSSL Dockerfiles. I tried changing this to the latest relevant tag I saw (AWS-LC-FIPS-1.0.2), but the Docker image failed to build because (I assume) that tag is a bit out of date. Can you tag the working commit on `main` and pin that tag here? Then I can build the image and set up a CI build that runs within it.

samuel40791765: Sorry for the confusion, the `AWS-LC-FIPS-1.0.2` branch points to a FIPS version release, which points to a branch of an older commit of the repo that's still in the process of getting FIPS validated.
I've set the docker image tag to `v1.1.0`, which points to a newer commit should work. Just so you know, we'll also be setting up an encryption-sdk CI in AWS-LC, so we can verify that all future versions of AWS-LC will build with ESDK :)

ajewellamz (dismissed):

ajewellamz (approved):

PR #758, opened 2022-08-30
fix: kms delay

ajewellamz:
*Description of changes:*

Delay 0.1 seconds before each KMS request, in hopes that our intermittent failures can be solves by throttling.

# Check any applicable:

### DIFF
--- a/aws-encryption-sdk-cpp/tests/test_vectors/static_test_vectors.cpp
+++ b/aws-encryption-sdk-cpp/tests/test_vectors/static_test_vectors.cpp
@@ -288,6 +288,7 @@ static aws_cryptosdk_keyring *keyring_for_master_key_spec(
             }
         }
     } else if (!cmp_jsonstr_with_cstr(key_type_obj, "aws-kms")) {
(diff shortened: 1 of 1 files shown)


aws-encryption-sdk-cpp/tests/test_vectors/static_test_vectors.cpp
```diff
@@ -288,6 +289,7 @@ static aws_cryptosdk_keyring *keyring_for_master_key_spec(
             }
         }
     } else if (!cmp_jsonstr_with_cstr(key_type_obj, "aws-kms")) {
+        usleep(100);
```
alex-chew, line 292: nit: let's pull this value out into a constant

alex-chew: or better yet, since `keyring_for_master_key_spec` writes the key type into `test_type_idx`, we can move this delay logic into the caller in order to reduce duplication

alex-chew (approved):
LGTM

PR #765, opened 2023-03-06
test: upgrade awslc dependency

ajewellamz:
*Issue #, if available:*

*Description of changes:*

upgrade AWS-LC dependency

# Check any applicable:


texastony (dismissed):
Unfortunately, this is not `feat`.
It is just a `test`.
`codebuild/bin/install-shared-deps-awslc.sh` is not used in the ESDK-C's src,
but only when doing integration tests.

I am not even sure we direct customer's to it.

If you change the PR title to:
`test: upgrade awslc dependency`, I will approve.

texastony (approved):

PR #781, opened 2024-04-24
chore: use built-in SAT solver for CBMC proofs

tautschnig:
*Description of changes:*

Invoking an external solver is not necessarily cheaper in terms of CPU or memory usage.

# Check any applicable:
- n/a Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.


tautschnig:
With #780 and this one we have now reduced the time for running proofs from around 1 h 15 min down to 40 mins.

ajewellamz (dismissed):

PR #812, opened 2025-02-07
fix: CMake Fixes to Make it Compatible with Latest Aws-c-common

waahm7:
*Description of changes:*
Hi, I am from the Common Runtime team and we own https://github.com/awslabs/aws-c-common. We recently did some cleanup in CMake, in aws-c-common at https://github.com/awslabs/aws-c-common/pull/1178. This broke this repository because it was relying on aws-c-common to set `LIBRARY_DIRECTORY` and `RUNTIME_DIRECTORY`, which aws-c-common no longer does. This PR updates aws-encryption-sdk-c to now use the official CMake variables like aws-c-common.
Changes: 
- Respect `CMAKE_INSTALL_LIBDIR`, `CMAKE_INSTALL_BINDIR`, `CMAKE_INSTALL_INCLUDEDIR`
    - Use these variables directly, instead of defining our own (~LIBRARY_DIRECTORY~, ~RUNTIME_DIRECTORY~)
    - `include(GNUInstallDirs)` on all platforms
    - REASONING: This removes a bunch of boilerplate, and makes it obvious we should defer to CMake for install directory names. I assume the original author thought `GNUInstallDirs` wasn't safe for Windows and Apple, but despite its name, it's fine. [Professional CMake](https://crascit.com/professional-cmake/release-notes/12th-edition/) says:
        > CMake’s GNUInstallDirs module provides a very convenient way to use a standard directory layout. ... it also provides various other standard locations that conform to both GNU coding standards and the FHS. ... the layout can even be used for Windows deployments

We can improve the CMake further like we did in aws-c-common, which will also fix #811, but I have decided to keep this PR simple, so that it is easier to review and merge.

- macOS-11 is no longer available, use the minimum macOS-13 github runner.
- Use macos-latest-large which is the Intel based runner instead of macos-latest which is now arm based.

# Check any applicable:


graebm (approved):
LGTM (I had made the changes to aws-c-common)

graebm (approved):

ajewellamz (approved):
Looks good to me.

PR #817, opened 2025-03-05
chore: add support for new style Dafny test vectors

ajewellamz:
*Issue #, if available:*

*Description of changes:*

# Check any applicable:


texastony (commented):
This is awesome.
I wish I was as Groovy as @ajewellamz .

I glanced over most of the C code,
by-passing the base64 logic,
as it is essentially a dependency we are in-lining.

I was super relieved when I realized that most of the 500 files were an expanded zip of test vectors from another implementation, Dafny.

I was terrified that @ajewellamz had written 500 files of C,
and would hate me for the rest of his days.

tests/TestVectors/Makefile
mahnushm, line 1: Some small changes: 
1) Using *.cpp in the compilation step can lead to unexpected behavior if unnecessary files exist in the directory. It is better to explicitly list source files.

2) No clean target exists to remove compiled binaries and temporary files.  Add: clean:
    rm -f test_vectors

3) If ./test_vectors fails, the Makefile continues execution. Add: decrypt_dafny: test_vectors
    ./test_vectors decrypt --manifest-path ./from-dafny --manifest-name decrypt-manifest.json || exit 1

tests/TestVectors/do_decrypt.cpp
mahnushm, line 1: 1) Line 32. What happens to keyring if aws_cryptosdk_session_new_from_keyring_2 fails? 

2) read_file(test.ciphertext, dir) and read_file(test.result, dir) are assumed to succeed and are not checked for errors.

tests/TestVectors/do_encrypt.cpp
```diff
@@ -0,0 +1,210 @@
...
+        printf("Failed to encrypt: %d %s\n", code, ERROR);
+        return results.bump(Result::Fail, test);
+    }
+    aws_cryptosdk_session_destroy(session);
+    ciphertext_result.resize(ciphertext_result_len);
+    results.bump(Result::Pass, test);
+    return MakeDecryptJson(test, ciphertext_result, dir);
+}
+
+static json RunEncryptTestK(
```
mahnushm, line 148: Very similar to RunEncryptTest(), can we just pass one extra parameter (keyring) and  handle it inside the function to reduce code duplication?

mahnushm (dismissed):

tests/TestVectors/do_decrypt.cpp
ajewellamz: read_file throws an exception if an error occurs. There is no error to be checked.
The problem did exist in write_file, which now also throws exceptions appropriately.

tests/TestVectors/do_encrypt.cpp
ajewellamz: I don't understand, that's exactly what I did.
There is no code duplication. If I have a keyring, I call RunEncryptTestK which creates a cmm, and calls RunEncryptTest.
Which of the two lines of RunEncryptTestK feel like duplication?

mahnushm (approved):
All is good, Thanks!

PR #825, opened 2025-07-03
chore(CI): clean up gha and unused files

josecorella:
*Issue #, if available:*

*Description of changes:*

# Check any applicable:

### DIFF
--- a/.github/workflows/daily_ci.yml
+++ b/.github/workflows/daily_ci.yml
@@ -1,0 +1,11 @@
+# This workflow runs every weekday at 15:00 UTC (8AM PDT)
+name: Daily CI
+
+on:
+  schedule:
+    - cron: "00 15 * * 1-5"
+
+jobs:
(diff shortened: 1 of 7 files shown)


.github/workflows/daily_ci.yml
```diff
@@ -0,0 +1,12 @@
+# This workflow runs every weekday at 15:00 UTC (8AM PDT)
```
lucasmcdonald3, line 1: This is fine and an improvement over the current state, but any reason not to include the other CI?

lucasmcdonald3 (approved):

.github/workflows/daily_ci.yml
josecorella: currently, this is the only job that is triggered daily, we could add the other ones; except osx - since you need to physically click run for those. we could add the others, I just wanted to get back to what we had

